Issue #86, opened 2025-01-29
00345 Add monitoring and health check features - phase II


FionaDyalog:
Mantis 21912 details the changes needed for this project (https://mantis.dyalog.com/view.php?id=21912)
Documents impacted in this project:
• Language Reference Guide
• Release Notes

PR #182, opened 2025-03-06
86 uuid generator

DyalogRichard:
References #86

### DIFF
--- a/language-reference-guide/docs/the-i-beam-operator/generate-uuid.md
+++ b/language-reference-guide/docs/the-i-beam-operator/generate-uuid.md
@@ -2,16 +2,14 @@
 
 This function generates a [UUID](https://en.wikipedia.org/wiki/Universally_unique_identifier) (Universally Unique IDentifier) according to the [RFC 9562](https://datatracker.ietf.org/doc/html/rfc9562) specification.
 
-A UUID is a label used to uniquely identify objects in computer systems and does not depend on a central registration authority or coordination between the parties generating them. UUIDs are also known as GUIDs (Globally Unique IDentifiers).
+A UUID is a label that uniquely identifies objects in computer systems; it does not depend on a central registration authority or co-ordination between the parties generating them. UUIDs are also known as GUIDs (Globally Unique IDentifiers).
 
-`Y` specifies the UUID Version required.
+`Y` specifies the UUID version required. Supported values of `Y` are:
 
-|`Y`| Version                                                       |
-|---|---------------------------------------------------------------|
-| 4 | UUIDv4: random values                                         |
-| 7 | UUIDv7: time-ordered (based on Unix Epoch) plus random values |
-
-Other values are either unsupported, undefined or reserved.
+|`Y`| Version                                                                                       |
+|---|-----------------------------------------------------------------------------------------------|
+| 4 | UUIDv4: random values                                                                         |
+| 7 | UUIDv7: a combination of values that are time-ordered (based on Unix Epoch) and random values |
 
 The result `R` is a vector containing the generated 36-character UUID.
 
--- a/language-reference-guide/docs/the-i-beam-operator/generate-uuid.md
+++ b/language-reference-guide/docs/the-i-beam-operator/generate-uuid.md
@@ -1,4 +1,4 @@
-<h1 class="heading"><span class="name">UUID Generator</span> <span class="command">R←120⌶Y</span></h1>
+<h1 class="heading"><span class="name">Generate UUID</span> <span class="command">R←120⌶Y</span></h1>
 
 This function generates a [UUID](https://en.wikipedia.org/wiki/Universally_unique_identifier) (Universally Unique IDentifier) according to the [RFC 9562](https://datatracker.ietf.org/doc/html/rfc9562) specification.
 
--- a/language-reference-guide/docs/the-i-beam-operator/i-beam.md
+++ b/language-reference-guide/docs/the-i-beam-operator/i-beam.md
@@ -31,7 +31,7 @@ The column labelled *O/S* indicates if a function applies only on Windows (W), o
 |`8`    |[Inverted Table Index-of](./inverted-table-index-of.md)                               |&nbsp;|
 |`43`   |[Monadic Operator Generator](./monadic-operator-generator.md)                         |&nbsp;|
 |`85`   |[Execute Expression](./execute-expression.md)                                         |&nbsp;|
-|`120`  |[UUID Generator](./uuid-generator.md)                                                 |&nbsp;|
+|`120`  |[Generate UUID](./generate-uuid.md)                                                   |&nbsp;|
 |`127`  |[Overwrite Free Pockets](./overwrite-free-pockets.md)                                 |&nbsp;|
 |`180`  |[Canonical Representation](./canonical-representation.md)                             |&nbsp;|
 |`181`  |[Unsqueezed Type](./unsqueezed-type.md)                                               |&nbsp;|
--- a/language-reference-guide/docs/the-i-beam-operator/i-beam.md
+++ b/language-reference-guide/docs/the-i-beam-operator/i-beam.md
@@ -31,6 +31,7 @@ The column labelled *O/S* indicates if a function applies only on Windows (W), o
 |`8`    |[Inverted Table Index-of](./inverted-table-index-of.md)                               |&nbsp;|
 |`43`   |[Monadic Operator Generator](./monadic-operator-generator.md)                         |&nbsp;|
 |`85`   |[Execute Expression](./execute-expression.md)                                         |&nbsp;|
+|`120`  |[UUID Generator](./uuid-generator.md)                                                 |&nbsp;|
 |`127`  |[Overwrite Free Pockets](./overwrite-free-pockets.md)                                 |&nbsp;|
 |`180`  |[Canonical Representation](./canonical-representation.md)                             |&nbsp;|
 |`181`  |[Unsqueezed Type](./unsqueezed-type.md)                                               |&nbsp;|
--- a/language-reference-guide/docs/the-i-beam-operator/uuid-generator.md
+++ b/language-reference-guide/docs/the-i-beam-operator/uuid-generator.md
@@ -1,0 +1,11 @@
+<h1 class="heading"><span class="name">UUID Generator</span> <span class="command">R←120⌶Y</span></h1>
+
+`Y` is the UUID version, and must be 4.
+
+The result `R` is a character vector containing a [RFC 9562](https://datatracker.ietf.org/doc/html/rfc9562) UUID of the specified version.
+
+<h2 class="example">Example</h2>
+```apl
+      120⌶4
+32cd549f-eb33-4457-bf45-babf26dc2b53
+```
--- a/language-reference-guide/docs/the-i-beam-operator/uuid-generator.md
+++ b/language-reference-guide/docs/the-i-beam-operator/uuid-generator.md
@@ -1,8 +1,17 @@
 <h1 class="heading"><span class="name">UUID Generator</span> <span class="command">R←120⌶Y</span></h1>
 
-`Y` is the UUID version, and must be 4.
+This function generates a UUID (Universally Unique IDentifier) according to the [RFC 9562](https://datatracker.ietf.org/doc/html/rfc9562) specification. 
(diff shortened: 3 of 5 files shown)


language-reference-guide/docs/the-i-beam-operator/uuid-generator.md
```diff
@@ -0,0 +1,11 @@
+<h1 class="heading"><span class="name">UUID Generator</span> <span class="command">R←120⌶Y</span></h1>
+
+`Y` is the UUID version, and must be 4.
```
FionaDyalog, line 3: Suggested rewrite using one of the following:

`Y` is the UUID (Universally Unique IDentifier – also known in Microsoft Windows as a GUID, Globally Unique Identifier) version. Only type 4 is supported, therefore `Y` must be 4.

`Y` is the UUID (Universally Unique IDentifier) version. Only type 4 is supported, therefore `Y` must be 4.

[Aside: Convention is to say "Universally Unique IDentifier (UUID)", but I don't think in this case that is the most useful way of expressing it!]

language-reference-guide/docs/the-i-beam-operator/uuid-generator.md
```diff
@@ -0,0 +1,11 @@
...
+
+The result `R` is a character vector containing a [RFC 9562](https://datatracker.ietf.org/doc/html/rfc9562) UUID of the specified version.
```
FionaDyalog, line 5: Suggested rewrite:

The result `R` is a character vector containing an [RFC 9562](https://datatracker.ietf.org/doc/html/rfc9562) UUID of the specified version (4).

DyalogRichard:
I decided to re-enable UUID Version 7, so the docs have changed to reflect that.

FionaDyalog (approved):

language-reference-guide/docs/the-i-beam-operator/uuid-generator.md
```diff
@@ -1,8 +1,17 @@
...
+`Y` specifies the UUID Version required, from the table below. The result `R` is a vector containing the generated 36-character UUID.
+
+|Version|Description|
+|---|--------------------------------------|
+| 4 | Randomly or pseudorandomly generated |
+| 7 | Unix Epoch time-based                |
+
+Other values are either unsupported, undefined or reserved.
+
+UUIDs are also known as GUIDs (Globally Unique Identifiers); the terms are equivalent.
```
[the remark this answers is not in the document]
DyalogRichard: I am quite keen to leave the explanation about GUIDs being the same as UUIDs at the bottom. It's not our job to explain general terminology and certainly not put it in the lead - what is important is what pertains to Dyalog. However, it's probably worth putting GUID in there somewhere - if only to ensure that anyone who searches for that term finds it.

language-reference-guide/docs/the-i-beam-operator/uuid-generator.md
```diff
@@ -0,0 +1,20 @@
+<h1 class="heading"><span class="name">UUID Generator</span> <span class="command">R←120⌶Y</span></h1>
+
+This function generates a UUID (Universally Unique IDentifier) according to the [RFC 9562](https://datatracker.ietf.org/doc/html/rfc9562) specification. 
+
+`Y` specifies the UUID Version required, from the table below. The result `R` is a vector containing the generated 36-character UUID.
+
+|Version|Description|
+|---|--------------------------------------|
+| 4 | Randomly or pseudorandomly generated |
+| 7 | Unix Epoch time-based                |
```
abrudz, line 10: This is also random. Wikipedia calls 4 "random" and 7 "timestamp and random". The fact that 7 is based on the Unix Epoch is not really important, unless we include how to extract the time from UUIDv7: `⊃12 ¯1⎕DT 16⊥12↑16~⍨⎕IO-⍨(⎕D,⎕A)⍳1∘⎕C` (Maybe a future enhancement to the I-beam?)

language-reference-guide/docs/the-i-beam-operator/uuid-generator.md
```diff
@@ -0,0 +1,20 @@
+<h1 class="heading"><span class="name">UUID Generator</span> <span class="command">R←120⌶Y</span></h1>
+
+This function generates a UUID (Universally Unique IDentifier) according to the [RFC 9562](https://datatracker.ietf.org/doc/html/rfc9562) specification. 
+
+`Y` specifies the UUID Version required, from the table below. The result `R` is a vector containing the generated 36-character UUID.
```
abrudz, line 5: Why can `Y` not be an array of `4`s and `7`s? (If made to be so, then consider what should happen in the scalar case.) Note that `120⌶¨` is problematic on an empty array.

language-reference-guide/docs/the-i-beam-operator/uuid-generator.md
```diff
@@ -0,0 +1,20 @@
...
+
+|Version|Description|
+|---|--------------------------------------|
+| 4 | Randomly or pseudorandomly generated |
+| 7 | Unix Epoch time-based                |
+
+Other values are either unsupported, undefined or reserved.
+
+UUIDs are also known as GUIDs (Globally Unique IDentifiers); the terms are equivalent.
```
abrudz, line 14: Maybe add that when called a GUID, the UUID is usually enclosed in braces, i.e. `guid←'{',(120⌶4),'}'`

abrudz (changes requested):
I think the name should be "Generate UUID" as that reads nicely:
`120⌶7` is read "generate UUID 7"

language-reference-guide/docs/the-i-beam-operator/uuid-generator.md
DyalogRichard: I have been trying to figure out exactly how MS handle GUIDs. It is complicated because there are lots of places you can generate them but looking at .NET in particular:

* A [standard GUID](https://learn.microsoft.com/en-us/dotnet/api/system.guid.newguid?view=net-9.0) is Version 4, and its default display form has embedded '-' chars and no braces.
* It is also possible to generate [Version 7 GUIDs](https://learn.microsoft.com/en-us/dotnet/api/system.guid.createversion7?view=net-9.0); there is no sign of any other type.
* A GUID object can be formatted in one of [five ways](https://learn.microsoft.com/en-us/dotnet/api/system.guid.tostring?view=net-9.0) which includes wrapped in braces but by default is without.

I wonder if a later extension to our implementation (or now, it's not too late) should add an optional left argument to support those five formats?

language-reference-guide/docs/the-i-beam-operator/uuid-generator.md
DyalogRichard: A possible extension for the future. However, I backed off from this before when considering the implications of providing/encouraging a rapid-fire mechanism of generating UUIDs which are based on a clock. Should the time element be the same in them all or slightly different? If different, should the generation rate be controllable?  Could the similarity in UUIDs lead to any kind of security issue? At least if the user generates the UUIDs separately they can make these decisions and we don't end up with egg or our face or having to implement yet more options. Anyway, *this* is not the place for this discussion - I'll add it to the project plan.

language-reference-guide/docs/the-i-beam-operator/uuid-generator.md
```diff
@@ -0,0 +1,20 @@
+<h1 class="heading"><span class="name">UUID Generator</span> <span class="command">R←120⌶Y</span></h1>
+
+This function generates a UUID (Universally Unique IDentifier) according to the [RFC 9562](https://datatracker.ietf.org/doc/html/rfc9562) specification.
```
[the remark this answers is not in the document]
DyalogRichard: I have added the suggested text. With that longer introduction it also makes sense to move the explanation that GUIDs are equivalent there, so I have done that also.

DyalogRichard:
> I think the name should be "Generate UUID" as that reads nicely: `120⌶7` is read "generate UUID 7"

I will leave that decision to Fiona as she is reviewing all of the names.

language-reference-guide/docs/the-i-beam-operator/uuid-generator.md
DyalogRichard: Updated to describe as "random values" and "time-ordered (based on Unix Epoch) plus random values"

language-reference-guide/docs/the-i-beam-operator/uuid-generator.md
abrudz: Yes, I think adding those 5 formats is a good idea, but [Wikipedia also lists a couple of additional formats](https://en.wikipedia.org/wiki/Universally_unique_identifier#Textual_representation):
* `0x00000000000000000000000000000000`
* `00000000000000000000000000000000h`
That said, APL is a programming language, and all these formats, except for `X` are trivial to compose oneself. And I've never seen the `X` format used anywhere.

FionaDyalog:
> > I think the name should be "Generate UUID" as that reads nicely: `120⌶7` is read "generate UUID 7"
> 
> I will leave that decision to Fiona as she is reviewing all of the names.

Yes, I am happy with that change. Please ensure the file name and ToC is amended for this; I will update the Cheat Sheet and Release Notes.

DyalogRichard:
> > > I think the name should be "Generate UUID" as that reads nicely: `120⌶7` is read "generate UUID 7"
> > 
> > 
> > I will leave that decision to Fiona as she is reviewing all of the names.
> 
> Yes, I am happy with that change. Please ensure the file name and ToC is amended for this; I will update the Cheat Sheet and Release Notes.

Done.

language-reference-guide/docs/the-i-beam-operator/uuid-generator.md
DyalogRichard: Ok, I'll leave this to be a possible future extension. I think the MS docs illustrate that { ... } is one possible representation, but not only is it only one of several, it is not the default.

abrudz (approved):

abrudz (approved):

language-reference-guide/docs/the-i-beam-operator/i-beam.md
```diff
@@ -31,6 +31,7 @@ The column labelled *O/S* indicates if a function applies only on Windows (W), o
 |`8`    |[Inverted Table Index-of](./inverted-table-index-of.md)                               |&nbsp;|
 |`43`   |[Monadic Operator Generator](./monadic-operator-generator.md)                         |&nbsp;|
 |`85`   |[Execute Expression](./execute-expression.md)                                         |&nbsp;|
+|`120`  |[UUID Generator](./uuid-generator.md)                                                 |&nbsp;|
```
abrudz, line 34: Needs updated title and URL.

abrudz (changes requested):

language-reference-guide/docs/the-i-beam-operator/i-beam.md
DyalogRichard: Good catch

language-reference-guide/docs/the-i-beam-operator/generate-uuid.md
```diff
@@ -0,0 +1,22 @@
+<h1 class="heading"><span class="name">Generate UUID</span> <span class="command">R←120⌶Y</span></h1>
+
+This function generates a [UUID](https://en.wikipedia.org/wiki/Universally_unique_identifier) (Universally Unique IDentifier) according to the [RFC 9562](https://datatracker.ietf.org/doc/html/rfc9562) specification.
+
+A UUID is a label used to uniquely identify objects in computer systems and does not depend on a central registration authority or coordination between the parties generating them. UUIDs are also known as GUIDs (Globally Unique IDentifiers).
```
FionaDyalog, line 5: Suggested rewrite (from line 5 down to the example):

A UUID is a label that uniquely identifies objects in computer systems; it does not depend on a central registration authority or co-ordination between the parties generating them. UUIDs are also known as GUIDs (Globally Unique IDentifiers).

`Y` specifies the UUID version required. Supported values of `Y` are:

|`Y`| Version                                                       |
|---|---------------------------------------------------------------|
| 4 | UUIDv4: random values                                         |
| 7 | UUIDv7: a combination of values that are time-ordered (based on Unix Epoch) and random values |

The result `R` is a vector containing the generated 36-character UUID.

FionaDyalog (approved):

PR #189, opened 2025-03-09
105 array notation with no XPLookAndFeel configuration parameter

5jt:
WORK IN PROGRESS

This branch assumes configuration parameter `XPLookAndFeel` does **not** survive in v20.0.

It removes reference to Native Look and Feel from the Session Toolbars article and also from the Editor article, which was **not* part of the `105-array-notation` agenda. (No other work done on the Editor article.)

It also removes folder `img/nlf-disabled/` as the edited MDs contain no other references to its images.

**Discard this branch** if `XPLookAndFeel` survives in v20.0.


5jt:
Per Fiona: config parameter `XPLookAndFeel` survives in V20.

Issue #45, opened 2025-01-06
Dyalog v20.0: Mantis 21315 – 00333 An improved version of ⎕SH


FionaDyalog:
Mantis 21315 details the changes needed for this project (https://mantis.dyalog.com/view.php?id=21315)

Documents impacted in this project:
- Language Reference Guide
- Release Notes

xpqz:
[Adding Peter's comments here]

I have written the ⎕SHELL documentation now, and created a pull request with you as the reviewer.
I don't know what the process is going to be for selecting reviewers.

A couple of notes I wrote down while writing the docs:
* I cannot get the { .example} syntax to work. It simply renders as it is written.
* The guidelines show examples of a windows-specific info box, introduced by !!! windows, but when I render it, it just looks like any other info box.
* The current set of files and directories are sometimes confusing to navigate.

For example, I wanted to add a reference to Language Reference > System Functions > System Functions Categorised > Introduction, so I assumed the file to edit would be in the system-functions/system-functions-categorised directory, but that entire directory seems not to be used? It contains a bunch of files which are also found elsewhere: For example, there is a file for ⎕R called r.md, but it exists elsewhere as well:

- ./language-reference-guide/docs/system-functions/system-functions-categorised/tools-and-access-to-external-facilities/tools-and-access-to-external-utilities/r.md
- ./language-reference-guide/docs/system-functions/r.md

* Knowing where to add references to new content on old pages is tricky, but the list you put on the mantis issue helped. But adding a reference to ⎕SHELL from ⎕CMD for example, is tricky as ⎕CMD's docs are spread across multiple pages.

I find the order of navigation menus confusing in a number of cases:

* For I-beams (Language reference > The I-Beam operator), I think they are ordered by number, but the number isn't shown, and the ordering of the names makes no sense, so it is tricky to navigate.
* For system functions (Language reference > System functions > System functions A-Z), the ordering is alphabetical, but based on the spelling of the system functions themselves, not their "names". For example, ⎕NQ is after ⎕NPUT, but their names in the menu are "Enqueue Event" and "Write Text File".
* The same problem might exist elsewhere.

Also, I have chosen to call ⎕SHELL "Execute External Program" instead of "Shell", but that hasn't been discussed anywhere else.

xpqz:
I have raised a separate issue for the directory structure issue: #57

pmikkelsen:
Thanks, I have pushed a commit with the proposed changes.

[...]

PR #192, opened 2025-03-24
Changes to ⎕SHELL docs based on feedback from MK

pmikkelsen:
Add a few changes to the ⎕SHELL documentation, based on some feedback.

References #45

### DIFF
--- a/language-reference-guide/docs/system-functions/shell.md
+++ b/language-reference-guide/docs/system-functions/shell.md
@@ -6,6 +6,7 @@
 `R` contains information about the termination reason, along with output collected.
 
 The system function is cross-platform, but the command specification is inherently operating-system specific.
+Some shells, such as PowerShell, are cross-platform, which can help minimise the problem.
 
 ### Using the System's Shell
 When `Y` is a character vector, the contents are executed using the system's shell.
@@ -33,6 +34,8 @@ d-----        26-11-2024     12:56                ullu
 ### Direct Execution
 
 When `Y` is a vector of character vectors or an enclosed character vector, the first element of the array is treated as a program name, and the rest as individual arguments. The program is executed directly, without invoking the system's shell first. The program name can be an absolute path, a relative path, or the name of an executable in the current search path (operating-system specific). When the program is specified as a relative path, the name is resolved relative to the working directory, which can be set with the [`WorkingDir`](#workingdir) variant.
+
+In general, if the command does not depend on any shell features, both formats of `Y` can be used, but using direct execution should be preferred as it has lower overhead.
 
 ## Return Value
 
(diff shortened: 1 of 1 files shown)


language-reference-guide/docs/system-functions/shell.md
```diff
@@ -34,6 +36,8 @@ d-----        26-11-2024     12:56                ullu
 
 When `Y` is a vector of character vectors or an enclosed character vector, the first element of the array is treated as a program name, and the rest as individual arguments. The program is executed directly, without invoking the system's shell first. The program name can be an absolute path, a relative path, or the name of an executable in the current search path (operating-system specific). When the program is specified as a relative path, the name is resolved relative to the working directory, which can be set with the [`WorkingDir`](#workingdir) variant.
 
+In general, if the command does not depend on any shell features, both formats of `Y` can be used, but using direct execution should be preferred as it has lower overhead.
```
FionaDyalog, line 39: Suggested rewrite:

If the command does not depend on any shell features, then direct execution has lower overhead.

language-reference-guide/docs/system-functions/shell.md
```diff
@@ -5,7 +5,9 @@
 `Y` describes the command to execute. `Y` is specified as either a character vector or a vector of character vectors.
 `R` contains information about the termination reason, along with output collected.
 
-The system function is cross-platform, but the command specification is inherently operating-system specific.
+The system function is cross-platform, but the command specification will depend on the platform's shell, the program being invoked, and the expected format of the command line arguments.
```
FionaDyalog, line 8: Suggested rewrite:

Although `⎕SHELL` itself is cross-platform, the command specification depends on the platform's shell, the program being invoked, and the expected format of the command line arguments – `⎕SHELL` calls are only as cross-platform as the programs they execute. In practice, many command line tools that are useful when building cross-platform applications, such as `git` and `aws-cli`, *are* cross-platform tools, which makes this less likely to be an issue.

FionaDyalog (approved):

language-reference-guide/docs/system-functions/shell.md
mkromberg: Doesn't this suggest that there is only lower overhead *IF* the command does not depend on any shell features? But the overhead is always lower.

mkromberg: You cannot use direct execution if you require shell features. The point we're trying to make is that IF you don't require shell features, we recommend direct execution.

FionaDyalog: Suggested rewrite:

If the command does not depend on any shell features, then direct execution should be used as it has lower overhead.

Issue #199, opened 2025-03-27
⎕NMOVE: result of sample instruction is shown in different code segment than the related instruction


mbaas2:
see screenshot:

![Image](https://github.com/user-attachments/assets/2176662e-6206-46d6-910b-f3cba257a857)

PR #200, opened 2025-03-27
This addresses the issue by moving the result

mbaas2:
References: #199


bear8642:
Believe this have been separately fixed - see commits linked to #199

Issue #212, opened 2025-04-02
Changes related to renaming of primitive tracing


FionaDyalog:
Related to #89 

The name has changed from "primitive tracing to "inline tracing".

The following changes have been made to the software and need to be reflected in the documentation:
- The keyboard shortcut is now IT for “Inline Trace”
- The default shortcut on Microsoft Windows is now Ctrl+Alt+Enter (this works in both Classic and Unicode editions)
- On non-Windows, added Ctrl+Alt+Enter to those input tables which have Ctrl+Enter for TC
- On non-Windows, added Ctrl+’k’ to those input tables which have Ctrl ‘r’ for TC

PR #214, opened 2025-04-02
Update the nomenclature from TBT to Inline Tracing

xpqz:
Rename token by token (TBT) tracing to "Inline Tracing" 

- Command becomes "IT"
- Rename file

References: #212

### DIFF
--- a/windows-ui-guide/docs/tracer/inline-tracing.md
+++ b/windows-ui-guide/docs/tracer/inline-tracing.md
@@ -1,34 +1,36 @@
-<h1 class="heading"><span class="name">Trace Primitives</span></h1>
+<h1 class="heading"><span class="name">Inline Tracing</span></h1>
 
-The ability to trace primitives is an extension to the Tracer that allows you to step through the execution of individual primitives within expressions, examining intermediate results and arguments of sub-expressions. It enables an in-depth inspection of complex expressions typed directly into the session, and can be used in conjunction with the traditional tracing mode to skip over lines you're not interested in and step through primitive-by-primitive in complex expressions where required.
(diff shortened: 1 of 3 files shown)


windows-ui-guide/docs/tracer/inline-tracing.md
```diff
@@ -1,34 +1,36 @@
...
+*Inline Tracing* is an extension to the Tracer that allows you to step through the execution of individual primitives within expressions, examining intermediate results and arguments of sub-expressions. It enables an in-depth inspection of complex expressions typed directly into the session, and can be used in conjunction with the traditional tracing mode to skip over lines you're not interested in and step through primitive-by-primitive in complex expressions where required.
 
 !!! note
-    _Trace Primitives_ is not an entirely accurate name: it is tracing with the (approximate) granularity of primitives, though it does stop on non-primitives.
+    _Inline Tracing_ is tracing with the (approximate) granularity of primitives, though it does stop on non-primitives, such as inlined dfns.
```
FionaDyalog, line 6: Suggested rewrite:

_Inline Tracing_ is tracing with the (approximate) granularity of primitives; it also stops on non-primitives, such as single-line dfns.

(I assume that this is what was meant by inlined dfns, if I have misunderstood please let me know!)

FionaDyalog (approved):
The changes for non-Windows platforms are not included here.

windows-ui-guide/docs/tracer/inline-tracing.md
xpqz: Not quite; the single-line aspect isn't important, it's that it's presented inline, as in not via a name. But thinking about it, the statement holds true for named dfns, too -- the inline tracing will stop on "MyFn" if "MyFn" is a function (dfn or not). Perhaps I should just say "user defined functions".

FionaDyalog: Sounbds good, but "user-defined" please.

Issue #230, opened 2025-04-09
First table under 1200⌶ (Format Date-time) in LR is missing its refs/footnotes


DyalogRichard:
The table under 1200⌶ which explains the formatting patterns should have three footnotes, and refs to those footnotes within the table (NB there are four refs to footnote 1). These are missing in 20.0.

Issue #225, opened 2025-04-08
1200 I-Beam formatting table conversion error


DyalogRichard:
There are some important leading spaces missing from the "Example" column of the table of formatting patterns.  For example,

Month | MM | 2 character numeric | _M | 3

there should be a space in front of the 3. The underscore in the pattern _M means pad 1 digit numbers with space rather than 0.

This is the case for all of the patterns of the form _X. It also applies to _t, which is incorrect in the 19.0 docs too (it shows the example as 08 for both tt and _t, but the example for tt should be 08 and the example for _t should be space-8).

xpqz:
This is an issue -- leading and trailling whitespace is not considered significant in a cell. Perhaps this should be a code block.

PR #231, opened 2025-04-09
Fix significant spacing in examples column of 1200I

xpqz:
The main table on the 1200⌶ had several errors:

- Significant leading spaces in the examples column for formatting options with a leading underscore was lost (also a problem in the v19 "old" docs) - added
- The example for "hours in twelve hour clock" for "tt" was missing a cell " 8" (note leading space) - added
- The footnotes could not be converted to appear in a HTML table easily using the mkdocs feature -- added as static section

References: #225 #230


DyalogRichard (approved):
Reviewed in person and all good. Addresses issues 225 and 230.

In the PDF file, the zero and space are not the same width so e.g. "08" above " 8" does not quite cause the "8"s to line up, but this is considered unimportant - the space is still present and visible.

Issue #227, opened 2025-04-08
)ED does not document Array Notation format


DyalogRichard:
In the Language Reference, neither ⎕ED or )ED include ⋄ (edit as Array Notation) in the symbol prefix table.

PR #295, opened 2025-06-05
Add ⋄ as prefix for editing array notation

bear8642:
Add ⋄ to the prefix tables for `)ED` and `⎕ED`

References #227


xpqz (approved):
+1

Issue #95, opened 2025-01-29
Mantis 21291 – ⎕FTIE/⎕FSTIE: New variant option


FionaDyalog:
Mantis 21291 is for a Dyalog v20.0 enhancement (https://mantis.dyalog.com/view.php?id=21291)

Documents impacted in this project:
•  Language Reference Guide
•  Release Notes

PR #297, opened 2025-06-06
⎕FTIE/⎕FSTIE: Add new variant option

bear8642:
As #95 states, document that ⎕FSTIE/⎕FTIE have a new Mode variant option in v20.0

### DIFF
--- a/language-reference-guide/docs/system-functions/fstie.md
+++ b/language-reference-guide/docs/system-functions/fstie.md
@@ -52,3 +52,19 @@ to:
 ```
 
 
+# Variant Options
+## Mode Variant
+
+Variant option ‘Mode’ specifies the intended use of the tie and can affect how and when errors are signalled.
+
+-|---
+R| File must only be be read. Any attempt to write to the file will fail.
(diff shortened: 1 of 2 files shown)


DyalogRichard:
Suggested rewrite:

# Variant Options
## Mode

The Mode variant option can be used to specify that, whilst tied, the file will be only be read, or that it must be possible to write to it. An error will be signalled if these conditions are not met.

Component files are not always writable - for example:

* The operating system permissions may not allow it.
* The file properties may not allow it.

If the use of the file is unspecified, a file that is not writable will be tied for read access if possible, and subsequent attempts to write to it will fail.

The Mode option may be used to explicitly specify how the file will be used: if for reading, the file will be tied for read access regardless of whether the file is writeable or not; if for writing, the tie will fail immediately if the file is not writable.

The options are as follows:

|Mode|Meaning|
|----|---
|`R` | Read. File is tied for read access and any subsequent attempt to write to it will fail.
|`U` | Unspecified (default). A file that is not writable will be tied for read access and any subsequent attempt to write to it will fail. 
|`W` | Write. A file that is not writable will fail to tie.

Mode is independent of any file access controls managed using an access matrix.

DyalogRichard (approved):

Issue #291, opened 2025-06-05
Mantis 22095 – ⎕VGET: add a note in the Case 2 block to point to how to simply do multiple names with same fallback


FionaDyalog:
Re https://mantis.dyalog.com/view.php?id=22095

This imapcts the following documents:
- Dyalog APL Language Reference Guide

PR #301, opened 2025-06-10
Tweaks to ⎕VGET example text

pmikkelsen:
References issue #291


dyaandys (approved):
I believe that multiple names with the same fallback will be the most common use case, so good to make it easy to find it in the documentation.

FionaDyalog (approved):
LGTM

Issue #309, opened 2025-06-10
Synonymise primitive functions


xpqz:
As some pages don't actually contain the primitive they're detailing as a token (e.g the page for match only contains `X≡Y`, so that is a "word" with a middle letter of ≡), searching for the primitive glyph doesn't always show the dyadic version.

Add a hidden search synonym.

PR #310, opened 2025-06-10
Add synonyms to the primitive functions and operator pages

xpqz:
Added hidden synonym div at the beginning of primitive function and operator pages to ensure that searching for the primitive glyph will result in both monadic and dyadic versions.


xpqz:
Closes: #309

Issue #314, opened 2025-06-11
Undangle dangling links


xpqz:
Still many dangling links reported:

- [ ] dotnet-framework-interface: 1
- [ ] earlier-release-notes: 3
- [ ] language-reference-guide: 55
- [ ] object-reference: 3
- [ ] programming-reference-guide: 17
- [ ] unix-installation-and-configuration-guide: 1
- [ ] unix-user-guide: 1
- [ ] windows-installation-and-configuration-guide: 7
- [ ] windows-ui-guide: 10

(not checked the `release-notes` as a WIP)

PR #318, opened 2025-06-11
dangling links undangled, part 2

xpqz:
Second batch of danglers undangled.

References: #314


FionaDyalog (approved):
I'm not confident reviewing the validity of the relative links, but other than that they look good!

Issue #68, opened 2025-01-14
Dyalog v20.0: Mantis 20331 – 00316 Performance improvements in membership and dyadic iota


FionaDyalog:
Mantis 21331 details the changes needed for this project (https://mantis.dyalog.com/view.php?id=21331)
Documents impacted in this project:
•  Language Reference Guide
•  Release Notes

PR #329, opened 2025-06-16
Remove ibeams for "Performance improvements in membership and dyadic iota"

bear8642:
As #68 states, remove the no longer relevant `⌶` beams


FionaDyalog (changes requested):
One further file needs to be updated to remove these two I-beams: language-reference-guide/print_mkdocs.yml

FionaDyalog (approved):
LGTM

PR #330, opened 2025-06-16
Ensure that search exclusion and synonyms are mutually exclusive

xpqz:
* Search-excluded pages should not have search synonyms.
* Ensure that anything added to pages with yml frontmatter must retain the yml at the front!
* Update utility scripts to encode these assertions.


FionaDyalog (approved):
I'm not the ideal reviewer for this as I'm yet fully au fait with Git/Markdown/MKDocs, but trust that Stefan's experience means that this is correct.

Issue #348, opened 2025-06-19
Mantis 21900 – Document that +/ goes from left to right


bear8642:
This relates to [Mantis 21900](https://mantis.dyalog.com/view.php?id=21900)

Documents impacted:

- Dyalog APL Language Reference Guide
- Dyalog Programming Reference Guide

PR #349, opened 2025-06-19
Document that some primative commutative reductions go left to right

bear8642:
As #348 states, some primative  commutative reductions goes from left to right.

Add a warning about this.

### DIFF
--- a/language-reference-guide/docs/primitive-operators/reduce.md
+++ b/language-reference-guide/docs/primitive-operators/reduce.md
@@ -31,6 +31,12 @@ Otherwise, if `f` is Catenate, `R` is `S⍴⊂0/⊃Y`. If `f` is Catenate First,
 
 Otherwise, `DOMAIN ERROR` is reported.
 
+!!! Warning
+	Due to historic implementation choices, `+` and `×` reductions are applied in the opposite order to what has been descibed above.
+    
+	That is, for `+/Y` and `×/Y`, the result `R` is:
+           `R ←→ ⊂(1⊃⌽Y)f(2⊃⌽Y)f......f(n⊃⌽Y)`
+
 Table: Identity Elements {: #IdentityElements }
 
 |Function|&nbsp;| Identity |


language-reference-guide/docs/primitive-operators/reduce.md
```diff
@@ -31,6 +31,12 @@ Otherwise, if `f` is Catenate, `R` is `S⍴⊂0/⊃Y`. If `f` is Catenate First,
 
 Otherwise, `DOMAIN ERROR` is reported.
 
+!!! Warning
+	Due to historic implementation choices, `+` and `×` reductions are applied in the opposite order to what has been descibed above.
+    
+	That is, for `+/Y` and `×/Y`, the result `R` is:
+           `R ←→ ⊂(1⊃⌽Y)f(2⊃⌽Y)f......f(n⊃⌽Y)`
+
```
FionaDyalog, line 39: Rewrite this:

!!! Warning "Warning"
    An early implementation decision that `+` and `×` reductions would not follow the right-to-left ISO standard implementation means that, in these two specific instances (`+/Y` and `×/Y`), the result `R` is:
```apl
      R ←→ ⊂(1⊃⌽Y)f(2⊃⌽Y)f......f(n⊃⌽Y)
```

FionaDyalog (approved):

language-reference-guide/docs/primitive-operators/reduce.md
DyalogRichard: IMO, more fundamental changes are needed here.

In my copious spare time I will suggest an alternative for your consideration.

Can you hold off on this one until I have done so?

DyalogRichard: I think we ought to explain this a bit more, if we're going to say anything at all. (I do question whether we need to at all.) I certainly wouldn't try constructing a complicated expression to illustrate the difference between right to left and left to right. This text needs some polish (Fiona?) but I would suggest if we're going to do this, go for something like this:

Whereas APL expressions evaluate right to left, from its first release Dyalog APL has evaluated + and × reductions in left to right (non-ISO) order because + and - are commutative and left to right evaluation was faster. However, the imprecise way in which which large and non-integral values can be stored does mean that there are some cases where left to right evaluation gives different results from right to left. For example, when ⎕DR is 645, 2*53 and 1+2*53 are equal within the limits of floating point precision, and therefore

(2*53)+1+(-2*53)

yields 0 when evaluated left to right, and 1 when evaluated right to left.

Issue #356, opened 2025-06-23
Changes for ⎕VGET/⎕VSET to reflect updated specification


pmikkelsen:
There has been some updates to the ⎕VGET/⎕VSET specification, which require documentation changes.
The project plan on the wiki has been updated (diff: https://wiki.bramley.dyalog.com/index.php?title=00354_Name_setter%2Fgetter_%28%22safe_execute%22%29&diff=63459&oldid=63312)

and the related mantis issues are:
* https://mantis.dyalog.com/view.php?id=22316
* https://mantis.dyalog.com/view.php?id=22317
* https://mantis.dyalog.com/view.php?id=22318
* https://mantis.dyalog.com/view.php?id=22319
* https://mantis.dyalog.com/view.php?id=22320
* https://mantis.dyalog.com/view.php?id=22322
* https://mantis.dyalog.com/view.php?id=22322
* https://mantis.dyalog.com/view.php?id=22323
* https://mantis.dyalog.com/view.php?id=22328

PR #357, opened 2025-06-23
Changes for ⎕VGET/⎕VSET

pmikkelsen:
References #356

### DIFF
--- a/language-reference-guide/docs/system-functions/vget.md
+++ b/language-reference-guide/docs/system-functions/vget.md
@@ -13,13 +13,15 @@
 * a vector of names – see [Case 2: Vector of Names](#case-2-vector-of-names).
 * a vector of nameclasses – see [Case 3: Nameclasses](#case-3-nameclasses).
 
-All specified names must have a nameclass of 0, 2, 8, or 9 in the source namespace(s). For more information on nameclasses, see [`⎕NC`](nc.md). If `Y` specifies a matrix or a vector of names, fallback values to use in cases where a name has a nameclass of 0 can also be specified to prevent a `VALUE ERROR` from being generated.
+All specified names must have a nameclass of 0, 2, 8, or 9 in the source namespace(s). For more information on nameclasses, see [`⎕NC`](nc.md). If `Y` specifies a matrix or a vector of names, fallback values to use in cases where a name has no value can also be specified to prevent a `VALUE ERROR` from being generated.
 
 If specified, `X` must be an array that identifies one or more source namespaces. This means that `X` must be one of:
 
 * a simple character scalar or vector identifying the name of a namespace.
 * a reference to a namespace.
 * an array in which each item is one of the above. If `X` refers to multiple namespaces, then `⎕VGET` processes each item of `X` in ravel order, using the entire right argument `Y`; this is equivalent to  `X ⎕VGET¨⊂Y`.
+
+When `X` is an empty array, the prototype of the empty result `R` depends on the specified fallback values, if specified. If the prototype of `X` is a namespace which can be instantiated, such as an instance of a class with a niladic or no constructors, the namespace is instantiated, and the prototype of the result depends on the values found there.
 
 The namespace(s) referenced must already exist, or a `VALUE ERROR` is generated.  
 
@@ -105,7 +107,7 @@ Names are specified as character vectors or scalars. `Y` must be one of the foll
 
 * a single name: `R` is the value of that name in the source namespace.
 * a single enclosed name: `R` is also the value of the name, but enclosed.
-* a single enclosed name-value pair, which is a two-element vector consisting of a character vector name and a fallback value for that name: `R` is the value of the name, or the fallback value in case the name has nameclass 0.
+* a single enclosed name-value pair, which is a two-element vector consisting of a character vector name and a fallback value for that name: `R` is the value of the name, or the fallback value in case the name has no value.
 * a nested vector where each item is either a name, or a name value pair: `R` is a vector with the same length as `Y`, with the values from the corresponding names, or fallback values.
 
 <h3 class="example">Examples</h3>
@@ -169,6 +171,8 @@ See [Case 1: Name Matrix](#case-1-name-matrix) for an example of multiple names 
 `Y` must be a numeric scalar or vector, where each item is a nameclass (see [Name Classification](nc.md)).
 
 If any of the numbers in `Y` are negative, the result `R` is a vector of name-value pairs, one for each existing name in the source namespace with a nameclass from `Y`. Otherwise, `R` is a 2-element nested vector, where the first element is a character matrix of names and the second element is a vector of values. In both cases, `R` is suitable as an argument for [`⎕VGET`](vget.md) and [`⎕VSET`](vset.md).
+
+In some cases, such as fields and variables with triggers attached, the nameclass of the symbol does not match the nameclass of the value. Therefore, [`⎕VGET`](vget.md) with a numeric right argument works on the nameclass of the values and not the names, except when the user explicitly requests nameclasses 2.1 or 2.2.
 
 <h3 class="example">Examples</h3>
 Name value pairs:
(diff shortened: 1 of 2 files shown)


language-reference-guide/docs/system-functions/vget.md
```diff
@@ -13,14 +13,16 @@
 * a vector of names – see [Case 2: Vector of Names](#case-2-vector-of-names).
 * a vector of nameclasses – see [Case 3: Nameclasses](#case-3-nameclasses).
 
-All specified names must have a nameclass of 0, 2, 8, or 9 in the source namespace(s). For more information on nameclasses, see [`⎕NC`](nc.md). If `Y` specifies a matrix or a vector of names, fallback values to use in cases where a name has a nameclass of 0 can also be specified to prevent a `VALUE ERROR` from being generated.
+All specified names must have a nameclass of 0, 2, 8, or 9 in the source namespace(s). For more information on nameclasses, see [`⎕NC`](nc.md). If `Y` specifies a matrix or a vector of names, fallback values to use in cases where a name has no value can also be specified to prevent a `VALUE ERROR` from being generated.
```
abrudz, line 16: I don't like the reference to `⎕NC` here, as `⎕VGET` disagrees with `⎕NC` about nameclasses. Instead, it should have its own description of the numbers.

language-reference-guide/docs/system-functions/vget.md
```diff
@@ -13,14 +13,16 @@
 * a vector of names – see [Case 2: Vector of Names](#case-2-vector-of-names).
 * a vector of nameclasses – see [Case 3: Nameclasses](#case-3-nameclasses).
```
abrudz, line 14: It bothers me a little that the statement "It must be one of the following:" isn't really true, because of the fallback values. Maybe it isn't a bug issue, but the phrasing "must contain one" would be a slightly better fit. best would be to spell it out fully:

* a matrix of names or a matrix of names and a value vector – see…
* a vector of names or name–value pairs – see…

abrudz (changes requested):

language-reference-guide/docs/system-functions/vget.md
```diff
@@ -170,6 +172,8 @@ See [Case 1: Name Matrix](#case-1-name-matrix) for an example of multiple names
 
 If any of the numbers in `Y` are negative, the result `R` is a vector of name-value pairs, one for each existing name in the source namespace with a nameclass from `Y`. Otherwise, `R` is a 2-element nested vector, where the first element is a character matrix of names and the second element is a vector of values. In both cases, `R` is suitable as an argument for [`⎕VGET`](vget.md) and [`⎕VSET`](vset.md).
 
+In some cases, such as fields and variables with triggers attached, the nameclass of the symbol does not match the nameclass of the value. Therefore, [`⎕VGET`](vget.md) with a numeric right argument works on the nameclass of the values and not the names, except when the user explicitly requests nameclasses 2.1 or 2.2.
```
mkromberg, line 175: The first sentence is incorrect, I think. The bug with triggers has been fixed. I suggest:

`⎕NC` always reports the names of fields in a class as having nameclass 2 (2.2 to be exact), even when the name has no value and you might expect 0, or the field is a namespace reference and you would expect 9. [`⎕VGET`](vget.md) with a right argument of 2 will only include fields that have values that are not references, while a right argument of 9 will include fields which are references. With a right argument of 2.2, `⎕VGET` will return all fields which are no underfined.

mkromberg: There is a use case for 2.2 (give me the fields and their values, regardless of whether they are really 2 or 9). There is no use case that I can imagine for 2.1 (give me all the variables of type 2, and include type 9 if they happen to have a trigger attached - this is arguably still a bug in the interpreter). So I would not do 2.1. I think it is unnecessary to also remove 2.2, but don't care enough to argue; we can always add it later.

abrudz: What was your suggestion for `2.1`? All non-ref non-fields? I think that makes sense, because then one can get the current behaviour (all non-fields with `2.1 9`).

abrudz: I'm happy with that.

mkromberg (approved):

language-reference-guide/docs/system-functions/vget.md
```diff
@@ -9,18 +9,20 @@
...
+All specified names must be either undefined, or have an array value in the source namespace(s). If `Y` specifies a matrix or a vector of names, fallback values to use in cases where a name has no value can also be specified to prevent a `VALUE ERROR` from being generated.
 
 If specified, `X` must be an array that identifies one or more source namespaces. This means that `X` must be one of:
 
 * a simple character scalar or vector identifying the name of a namespace.
 * a reference to a namespace.
 * an array in which each item is one of the above. If `X` refers to multiple namespaces, then `⎕VGET` processes each item of `X` in ravel order, using the entire right argument `Y`; this is equivalent to  `X ⎕VGET¨⊂Y`.
 
+When `X` is an empty array, the prototype of the empty result `R` depends on the specified fallback values, if specified. If the prototype of `X` is a namespace which can be instantiated, such as an instance of a class with a niladic or no constructors, the namespace is instantiated, and the prototype of the result depends on the values found there.
```
FionaDyalog, line 24: Suggested rewrite:

When `X` is an empty array, the prototype of the empty result `R` depends on the fallback values, if specified. If the prototype of `X` is a namespace that can be instantiated, such as an instance of a class with a niladic or no constructors, then the namespace is instantiated and the prototype of the result is determined by the values found within the instantiated namespace.

language-reference-guide/docs/system-functions/vget.md
```diff
@@ -170,6 +172,8 @@ See [Case 1: Name Matrix](#case-1-name-matrix) for an example of multiple names
 
 If any of the numbers in `Y` are negative, the result `R` is a vector of name-value pairs, one for each existing name in the source namespace with a nameclass from `Y`. Otherwise, `R` is a 2-element nested vector, where the first element is a character matrix of names and the second element is a vector of values. In both cases, `R` is suitable as an argument for [`⎕VGET`](vget.md) and [`⎕VSET`](vset.md).
 
+[`⎕NC`](nc.md) always reports the names of fields in a class as having nameclass 2 (2.2 to be exact), even when the name has no value and you might expect 0, or the field is a namespace reference and you would expect 9. [`⎕VGET`](vget.md) with a right argument of 2 will only include fields that have values that are not references, while a right argument of 9 will include fields which are references. With a right argument of 2.2, [`⎕VGET`](vget.md) will return all fields which are not undefined.
```
FionaDyalog, line 175: Suggested rewrite:

[`⎕NC`](nc.md) always reports the names of fields in a class as having nameclass `2` (`2.2` with the sub-class), even when the name has no value (might expect `0`) or the field is a namespace reference (might expect `9`). [`⎕VGET`](vget.md) with a right argument of `2` will only include fields that have values that are not references, while a right argument of `9` will include fields that are references. With a right argument of `2.2`, `⎕VGET` will return all fields that are not undefined.

language-reference-guide/docs/system-functions/vset.md
```diff
@@ -83,7 +83,7 @@ Multiple names, with a single value:
 ## Variant Option: Trigger
 
 The `Trigger` variant option specifies whether any [triggers](../../../programming-reference-guide/triggers/triggers) should be run for the modified variables in the target namespace that have triggers attached.
-The value must be a Boolean scalar. The default is 0, meaning that triggers are not run.
+The value must be a Boolean scalar. The default is 1, meaning that triggers are run.
```
FionaDyalog, line 86: 1 -> `1`

language-reference-guide/docs/system-functions/vset.md
```diff
@@ -98,12 +98,16 @@ The value must be a Boolean scalar. The default is 0, meaning that triggers are
...
+      ⍝ With the trigger option disabled, triggers are not run
       ⎕VSET⍠'Trigger' 0⊢('name1' 1) ('name2' 2) ('name3' 3)
 
       ⍝ With the trigger option enabled, triggers are run
-      ⎕VSET⍠'Trigger' 0⊢('name1' 1) ('name2' 2) ('name3' 3)
+      ⎕VSET⍠'Trigger' 1⊢('name1' 1) ('name2' 2) ('name3' 3)
+Running trigger for: name1
+Running trigger for: name3
+
+      ⍝ Without the trigger option, triggers are run as well
```
FionaDyalog, line 109: triggers are run as well
->
triggers are run (the default)

FionaDyalog (approved):

Issue #362, opened 2025-06-26
Incorrect font applied during conversion to MKDocs


FionaDyalog:
(reported by a user)

[Dyalog APL Language Reference Guide > Symbols](https://dyalog.github.io/documentation/latest/language-reference-guide/symbols/language-elements/)

Click on any of the symbols on this page to view the disambiguation pages. Many of these are very oddly fomatted, with heading fonts where one would expect body text.

Starting from the symbols page, every link needs to be clicked and the resulting page checked (and updated as necessary).

For example:

![Image](https://github.com/user-attachments/assets/12317912-89d0-47fa-95ae-e5ce3b383c75)

PR #364, opened 2025-06-30
Correct the font applied during conversion to MKDocs

bear8642:
As #362 states, the various Dyalog APL Language Reference Guide > Symbols sub-pages have heading fonts where body text was expected.

Correct these

### DIFF
--- a/language-reference-guide/docs/symbols/ampersand.md
+++ b/language-reference-guide/docs/symbols/ampersand.md
@@ -5,9 +5,9 @@ search:
 
 <h1 class="heading"><span class="name">Ampersand</span> <span class="command">&</span></h1>
 
-## Ampersand is a Monadic operator with an ambivalent operand
+Ampersand is a Monadic operator with an ambivalent operand
 
-## Operator Ampersand means
+Operator Ampersand means
 
 [Spawn](../primitive-operators/spawn.md)
 ```apl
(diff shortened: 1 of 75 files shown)


FionaDyalog (approved):
Approved (albeit without an extensive check), but would be nice to do the following:
- [x] Amend the title of the PR as "Correct font" is ambiguous..."Correct the font" would mitigate that!
- [x] Validate that all the glyph names match those in the [Cheat Sheet](https://docs.dyalog.com/latest/CheatSheet%20-%20Nomenclature%20-%20Functions%20and%20Operators.pdf) – they shold match the content of the "Glyph name" column, not the "Unicode glyph name" column.
- [ ] Validate that all the function/operator names also match thouse given in the above Cheat Sheet.
- [ ] Tweak the statements so that monadic/dydic are not capitalised unless they are the first word. Might be other things that fall into this category too.

If you don't have the capacity to do any/all of these, please let me know so that I can include them on a different issue that I'm raising about correcting the filenames. Thanks.

bear8642:
Re-requested review due to uncertainty about:
* Removal of the word "Sign" from the mathematical and comparison pages
* Some pages (`○` `,` `⌿` `∨` `∧`) have [Name (alias)] as the link text - should this be left alone?

Issue #385, opened 2025-07-16
Tables in Hold Statement documentation incorrectly rendered


bear8642:
The discussion of [`:Hold`](https://dyalog.github.io/documentation/20.0/programming-reference-guide/defined-functions-and-operators/traditional-functions-and-operators/control-structures/hold/) enabling deadlocks uses a table to show the code for two threads side by side.

Unfortunatly, the code is displayed all on one line

PR #387, opened 2025-07-16
Correct :Hold table rendering

bear8642:
As #385 states, `:Hold`'s uses tables to show code blocks adjacent to each other. 
Unfortunately the table and code blocks seem to not work together well.


xpqz (changes requested):
Don't do it like that...!

Keep the markdown table. Do line breaks inline with BR tags.

PR #399, opened 2025-07-23
Update docs deployment to include PDFs from the current SVN repository.

JasonRivers:
Git/SVN PDF work is done inside a container internally before being checked out on the remote server. This allows for a single job to update internal PDFs at the same time as the external website.


xpqz (approved):
+1 reviewed in person

Issue #378, opened 2025-07-08
Incorrect font in use in content


FionaDyalog:
Some references to APL code are not in the correct style.

For example, links to I-beams with https://dyalog.github.io/documentation/20.0/language-reference-guide/the-i-beam-operator/log-use-of-deprecated-features/

Check for this issue in all the new I-beams added for Dyalog v20.0.

DyalogRichard:
(Per "some references" - see https://dyalog.github.io/documentation/20.0/programming-reference-guide/deprecated-functionality/#deprecated-functionality-scheduled-for-removal for comparison, where there is also a link to 109⌶, but this time the font is correct.)

FionaDyalog:
All new I-beam pages for Dyalog v20.0 checked for code that ought to be in APL font, and entires that were not in this font have been amended so that they are.

PR #401, opened 2025-07-23
APL font applied to inline code

FionaDyalog:
Reference: #378


FionaDyalog:
All new I-beam pages for Dyalog v20.0 checked for code that ought to be in APL font, and entires that were not in this fonrt have been amended so that they are.

Closes #378

PR #414, opened 2025-07-31
Update Jenkins deployment to include readme files from Andys script.

JasonRivers:
Updated Jenkins deployment to include Dyalog readme files from SVN


dyaandys (approved):
>Looks good to me.  At some point I should review the combination of Jenkinsfile and scripts so that I define locations in only one place ..

xpqz (approved):
+1

Issue #429, opened 2025-08-20
Remove irrelevant example


FionaDyalog:
Re https://dyalog.github.io/documentation/20.0/language-reference-guide/primitive-operators/operator-syntax/

The examples in the introductory section at the top of this page include the following:

      ⎕NL 2
A
X
      ⎕EX¨↓⎕NL 2
      ⎕NL 2

This is not related to this page and should be removed.

PR #431, opened 2025-08-21
Remove irrelevant example

bear8642:
Remove irrelevant example from https://docs.dyalog.com/20.0/language-reference-guide/primitive-operators/operator-syntax as #429 shows


FionaDyalog (commented):
+1

Issue #415, opened 2025-08-05
Impossible to find keystroke to tab between windows in Unix User Guide


dyavc:
Impossible to find keystroke to tab between windows in Unix User Guide

https://dyalog.github.io/documentation/20.0/unix-user-guide/appendix-b-table-of-keycodes-putty/

I am trying to find out how to do the <TB> keystroke in a TTY dyalog on Linux.

Looking at the 20.0 docs webpage, it says:

Cmd represents the keystroke
CMD represents the keystrokes

I had to look at 19.0 docs to remind myself that Cmd is Ctrl+x

PR #432, opened 2025-08-21
Reformat Cmd and CMD keystroke definitions

bear8642:
With the move to markdown/html formating, reformat the Cmd and CMD keystroke definitions to reveal them
References #415


FionaDyalog (approved):
+1

PR #457, opened 2025-09-04
455 behind examples

FionaDyalog:
Enhanced the examples section.


language-reference-guide/docs/primitive-operators/behind.md
```diff
@@ -20,15 +20,163 @@ The derived function is equivalent to either `(f Y) g Y` or `(f X) g Y`, dependi
...
+Sort a vector by values in another vector:
+```apl
+      30 40 20 ⊂⍤⍋⍛⌷ 'Abe' 'Bea' 'Carl'
+Carl Abe Bea
+```
+
+Split a vector:
+```apl
+      ]Boxing on
+      Split←⊂⍤⍋⍛⌷
```
abrudz, line 172:
```suggestion
      Split←⊃⍛≠⊆⊢
```

pmikkelsen, line 172: This `Split` function seems to be the sort function from the example above, so the example doesn't produce the result shown here.

pmikkelsen (changes requested):

language-reference-guide/docs/primitive-operators/behind.md
```diff
@@ -20,15 +20,163 @@ The derived function is equivalent to either `(f Y) g Y` or `(f X) g Y`, dependi
...
+
+```
+
+Construct a Boolean mask from a length and the indices of the `1`s:
+```apl
+      10 ⍳⍛∊ 1 3 5
+1 0 1 0 1 0 0 0 0 0
+```
+
+Filter to return elements that match `0` in a Boolean mask:
```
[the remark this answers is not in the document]
FionaDyalog: How about the following compromise (want something that is in the same style as the others, which all start "construct", "take", "convert", etc.):

Filter to return elements that correspond to `0` in a Boolean mask – the opposite of _reduce_ (`/`), which returns elements that correspond to `1` in a Boolean mask.

language-reference-guide/docs/primitive-operators/behind.md
```diff
@@ -20,15 +20,163 @@ The derived function is equivalent to either `(f Y) g Y` or `(f X) g Y`, dependi
...
+ABCXYZ
+```
+
+Sort a vector by values in another vector:
+```apl
+      30 40 20 ⊂⍤⍋⍛⌷ 'Abe' 'Bea' 'Carl'
+Carl Abe Bea
+```
+
+Split a vector:
```
abrudz, line 169:
```suggestion
We can define a single text splitting function that can both be used monadically (in which case it uses the initial character as separator) and dyadically (in which case it uses the left argument as separator):
```

abrudz (changes requested):

Issue #471, opened 2025-09-09
Apply convention for system function names


FionaDyalog:
All references to system functions in the documentation should be fully capitalised.

For example, `⎕VGET` or `⎕NPUT`.

This impacts the following documents:

- [x] Dyalog APL Language Reference Guide
- [x] Dyalog Programming Reference Guide
- [x] Dyalog for Microsoft Windows Installation and Configuration Guide
- [x] Dyalog for Microsoft Windows UI Guide
- [x] Dyalog for Microsoft Windows Object Reference Guide
- [x] Dyalog for Microsoft Windows Interfaces Guide
- [x] Dyalog for Microsoft Windows .NET Framework Interface Guide
- [x] Dyalog for UNIX Installation and Configuration Guide
- [x] Dyalog for Microsoft Windows UI Guide

Any new documents brought across from the previous project should also be checked/amended as appropriate.

PR #476, opened 2025-09-10
Apply convention for system function names

bear8642:
As #471 states, system function names should be in uppercase.

The changes c2588d0b4f8...1843a90c583 were found programmatically (via`find  documentation -type f -iname '*.md' -exec grep -l '⎕[a-z]' {} \;`) and as such, other changes may be required.


bear8642:
Mixed case transformation by:
```
$ 9 sam -d `find . -type f -iname '*.md' -exec egrep -lI '⎕[a-zA-Z]+' {} \;`
```
then within [sam](https://en.wikipedia.org/wiki/Sam_(text_editor)):
```
X ,x/⎕[A-Za-z]+/y/⎕/y/[A-Z]/|tr a-z A-Z
X/'/w
q
```

xpqz (approved):
+1

Issue #478, opened 2025-09-10
"Partition" is missing its syntax


FionaDyalog:
https://dyalog.github.io/documentation/20.0/language-reference-guide/primitive-functions/partition/

The syntax has been lost in the migration proces and needs to be added.

cf https://help.dyalog.com/19.0/#Language/Primitive%20Functions/Partition.htm

PR #492, opened 2025-09-17
Add Partition's missing syntax text

bear8642:
As #478 states, the syntax text is missing for Partition. Restore it.


FionaDyalog (approved):
LGTM

Issue #547, opened 2025-10-15
Amend image links to point to "local" directrories


FionaDyalog:
Each document has its own images directory.

From an images perspective, documents should be self contained and not link to images in other documents (especially when the other document is the Release Notes, as those have a limited lifespan).

The following pages include links to images that are in the Dyalog v20.0 Release Notes – these need to be changed:

- programming-reference-guide/docs/introduction/arrays/display-of-arrays.md:
    -  release-notes/img/session_arraynotation.png
- windows-installation-and-configuration-guide/docs/configuration-parameters/aplan-for-output.md:
    - release-notes/img/session_arraynotation.png
- windows-installation-and-configuration-guide/docs/configuration-parameters/aplan-for-editor.md:
    - release-notes/img/object_arraynotation.png

The images should be copied into the appropriate directories and the links updated.

PR #558, opened 2025-10-15
Amend image links to point to local directories

bear8642:
As #547 states, the image paths need changing to enable the release notes to be removed as required.


FionaDyalog (approved):
LGTM

Issue #526, opened 2025-10-09
LangRef: Improve navigability and information related to glyphs


FionaDyalog:
The ToC currently includes a section with three pages:

<img width="361" height="189" alt="Image" src="https://github.com/user-attachments/assets/c7f3e68a-5b22-47fc-a6ff-6ca9ffb81929" />

This subsection and pages should be replaced with a single page called Glyphs.

The new page should comprise all information related to glyphs. It can include information from:

- /language-reference-guide/primitive-functions/primitive-functions-table/
- /language-reference-guide/primitive-operators/primitive-operators-table/
- /language-reference-guide/symbols/brackets/
- /language-reference-guide/symbols/special-symbols/

NOTE: As well as potentially breaking links within the documentation (which we can check for), the cross-references in help_urls.h (in SVN) will also need to be amended in line with the changes.

FionaDyalog:
Issue #442 should be considered when this is actioned. It can be closed when this is closed.

PR #561, opened 2025-10-15
Primitives that are not functions or operators moved to separate section

FionaDyalog:
New "other syntax" section created and populated with syntax previously erroneously listed as functions/operators.

- This fix will break some links elsewhere in this document – these will be corrected with #526 .
- No impact on help_urls.h


xpqz (commented):
Will this impact the help_urls.h? Also, I see no link changes; is that because there aren't any, or because that's to be a subsequent sweep?

xpqz:
+1

Issue #51, opened 2025-01-07
00385 Improve performance of creating a new process on AIX


FionaDyalog:
[Mantis 21682](https://mantis.dyalog.com/view.php?id=21682) details the changes needed for this project 

Documents impacted in this repository:
•  Language Reference Guide
•  Dyalog for UNIX Installation and Configuration Guide
•  Release Notes (done under issue #83)

PR #602, opened 2025-10-24
00385 improve performance of creating a new process on aix

bear8642:
Document the changes as #51 requires


FionaDyalog (commented):
Added feedback for the amended file.

Not yet approved as expecting the content of \language-reference-guide\docs\system-functions\shell.md to be amended to so will want to rereview whan that has been done.

PR #605, opened 2025-10-24
Link-checking a local build

xpqz:
Clarify that in order to link-check a local build, it needs to be up and running first.


FionaDyalog (approved):
LGTM

Issue #585, opened 2025-10-22
Housekeeping: remove files that are not required


FionaDyalog:
Check whether there are any .md files that are not included in any ToC (.yml file).

There are some for which this is valid, for example, disambiguation pages used by F1 help. Others should be checking individually and deleted if appropriate.

For example, \language-reference-guide\docs\primitive-functions\primitive-functions-tabulated.md can probably be deleted.

Suggestion: move all these into a separate directory then run the full link checker (see https://github.com/Dyalog/documentation/blob/main/tools/README.md) before deleting anything.

bear8642:
@FionaDyalog how has the orphaned pages list been generated?

FionaDyalog:
@bear8642 see the tools in https://github.com/Dyalog/documentation/tree/main/tools

bear8642:
This issue has been superseded by other housekeeping issues, such as #720 #722

PR #624, opened 2025-11-04
Remove redundant files

bear8642:
As #585 specifies, remove redundant files.

I've used the link checker to confirm that none of the files cause any issues there.


xpqz (commented):
I am not convinced this is safe -- specifically the changes to the disambiguation pages (the pages that only the CHM bundle sees).

FionaDyalog:
Let's come back to this after Dyalog v20.0 GA when we have the capacity to investigate this fully (including the CHM file generation).

PR #658, opened 2025-11-27
Split off Dyadic Execute into its own page

abrudz:
Also adds info on calling a function and expand dyadic example to illustrate both name and ref left arg.

### DIFF
--- a/language-reference-guide/docs/primitive-functions/execute.md
+++ b/language-reference-guide/docs/primitive-functions/execute.md
@@ -33,7 +33,7 @@ If the expression is an empty vector or a vector containing only blanks or one t
 
 
 
-If `Y` contains a branch expression, the branch is effected in the environment from which the Execute was invoked, and `⍎Y` does not return.
+If `Y` contains a branch expression, the branch is effected in the environment from which _execute_ was invoked, and `⍎Y` does not return.
 
 
 <h2 class="example">Examples</h2>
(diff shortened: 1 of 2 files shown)


language-reference-guide/docs/primitive-functions/execute.md
```diff
@@ -11,14 +11,18 @@ search:
 
 
 
-<h1 class="heading"><span class="name">Execute</span> <span class="command">R←{X}⍎Y</span></h1>
+<h1 class="heading"><span class="name">Execute</span> <span class="command">R←⍎Y</span></h1>
 
 !!! Warning
```
FionaDyalog:
```suggestion
!!! Warning "Warning"
    If the argument to _execute_ could include user input, then there is a risk to data and systems. To reduce this risk, a system function might be more appropriate than the _execute_ function. For example:

	* use [`⎕VGET`](../system-functions/vget.md) or [`⎕VSET`](../system-functions/vset.md to get or set the value of one or more variables named within one or more character vectors.
	* use [`⎕VFI`](../system-functions/vfi.md), [`⎕JSON`](../system-functions/json.md), or [`⎕CSV`](../system-functions/csv.md) to make numbers in text form into actual numbers.
    * use [`⎕OR`](../system-functions/or.md) to call a function by name; exact usage depends on valency. For example:
      * niladic – `(⎕OR fnName){⍺⍺}`
      * monadic – `(⎕OR fnName){⍺⍺ ⍵}YY`
      * dyadic – `X((⎕OR fnName){⍺ ⍺⍺ ⍵})Y`
```

Issue #630, opened 2025-11-05
References to tables not being displayed


FionaDyalog:
Re: https://docs.dyalog.com/20.0/language-reference-guide/primitive-operators/operators-summarised/

The references to the tables that should start this section are not being displayed. This should be corrected.

Rather than just amend this one page, need to investigate whether this is a wider issue that needs a fundamental change in the way we reference tables. Action after that will depend on the finding.

PR #671, opened 2025-12-05
Fix table refs by making minify plugin run last (#630)

xpqz:
- Minify plugin removes quotes around ids, which the table captioning regexes rely on to do forward references. 
- Solution: reorder plugins to minify after captioning
- Add explicit table captioning texts to all configs

### DIFF
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -30,14 +30,16 @@ plugins:
   - search
   - macros
   - monorepo
-  - minify:
-      minify_html: true
   - caption:
-      table: 
+      table:
         enable: true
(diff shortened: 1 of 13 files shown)


mkdocs.yml
```diff
@@ -30,14 +30,16 @@ plugins:
...
   - caption:
-      table: 
+      table:
         enable: true
         start_index: 1
         increment_index: 1
         position: bottom
+        reference_text: 'Table {index}'
+        caption_prefix: 'Table {index}:'
+  - minify:
```
xpqz: The crucial fix!

Issue #466, opened 2025-09-05
Documentation of LOG_FILE parameter shows a v19.0 path in sample


mbaas2:
...shouldn't that be updated?

https://dyalog.github.io/documentation/20.0//windows-installation-and-configuration-guide/configuration-parameters/log-file/

<img width="850" height="689" alt="Image" src="https://github.com/user-attachments/assets/293780ad-ae72-43bc-8917-6adeac86ba5a" />

DyalogRichard:
I suspect that is a wide-ranging and general problem, with a solution to use a substitution parameter for the "current" version.

I think there are some more specific problems here:

1. Users\<username>\Documents\
What on earth is that? It's not _the_ default - what's there looks kind'a Windowsy, but should have a drive letter and slash at the front, if so, and it's certainly nothing like what I see on Linux.

1. The setting is Log_File (following a long-standing, but incorrect, convention to capitalise in Pascal case), but appears instead as log_file in the navigation panel and on the page when the main title scrolls off the top.

PR #683, opened 2025-12-18
Improve LOG_FILE example

bear8642:
As #466 discusses, the example could be improved...


DyalogRichard:
It's very unlikely the default location "is" `C:\Users\<username>\Documents\..." - it does not *have* to be on drive C, it is usually under One Drive these days, and non-English Windows possibly doesn't use the names Users or Documents. I think this bit should be shown to come from the system somehow and only spelled out in the example.

DyalogRichard (changes requested):
We're not really going to change the version number in the "for example" every release, are we? It's an *example*; if we're changing stuff here anyway by all means update it at that point, but #466 ("Documentation of LOG_FILE parameter shows a v19.0 path in sample") should be rejected. 

Notwithstanding that, it has highlighted a different problem - the default is *not* "C:\Users\<username>\Documents\" - see comments elsewhere. Just adding a drive letter misses the point; to fix that the path should all be abstracted out somehow.

bear8642:
> It's very unlikely the default location "is" `C:\Users<username>\Documents..." - it does not _have_ to be on drive C, it is usually under One Drive these days [...] I think this bit should be shown to come from the system somehow and only spelled out in the example.

I'm happy to make it more obvious that it's an example LOG_FILE path.
Whilst a OneDrive file path is likely, it doesn't have to be and indeed isn't on my machine.

windows-installation-and-configuration-guide/docs/configuration-parameters/log-file.md
DyalogRichard: The current version of the file does not render correctly - the backslashes appear to be getting interpreted as escapes, for example. What is rendered is nonsense:

The default is C:\Users\Documents\Dyalog APL- . <Unicode|Classic> Files\default*.dlf_, for example, C:\Users\Bob\Documents\Dyalog APL-64 20.0 Unicode Files\default*.dlf_

The new text "If you have more than one instance of Dyalog installed..." made my head spin. John needs to look at that.

The text still says that the user's Documents directory **is** "C:\Users\\\<username\>\\Documents". In the general case it is not.

DyalogRichard (approved):
Two very minor typos - approval subject to those being fixed:

[1] "In this situation, at start-up, Dyalog attempts to open, and then locks, a file". To be gramattically correct, remove the "s" from the end of "locks".

[2] The default Log_File setting should have an underscore after "default" (i.e. "...\default_*.dlfx".

(Technically, the default Log_File setting should not be shown as "-&lt;bits&gt;" either, because on 32-bit neither the "-" or "32" is present, but that's too complicated and too unimportant to fix, I reckon.)

Issue #689, opened 2026-01-02
Reduce First N-Wise links to Reduce First rather than Reduce N-Wise


asherbhs:
**Page on which the error exists**
https://docs.dyalog.com/20.0/language-reference-guide/primitive-operators/reduce-first/reduce-first-n-wise/

**Version of the documentation in which you are reporting the error**
2025-12-04 main:a0aa8dc844

**Describe the error**
Incorrect link: 'See Reduce N-Wise' links to Reduce First rather than Reduce N-Wise

PR #692, opened 2026-01-08
Reduce First N-Wise links to Reduce First rather than Reduce N-Wise

bear8642:
As #689 states, Reduce First N-Wise incorrectly links to Reduce First rather than Reduce N-Wise


FionaDyalog (approved):
LGTM

Issue #708, opened 2026-02-04
Ensure that system functions that reference another system command use a link


bear8642:
**Page on which the error exists**
[State Indicator & Name List](https://docs.dyalog.com/20.0/language-reference-guide/system-commands/sinl/)
[Load without Latent Expression](https://docs.dyalog.com/20.0/language-reference-guide/system-commands/xload/)
[List Global Namespaces](https://docs.dyalog.com/20.0/language-reference-guide/system-commands/obs/)
[Load Workspace](https://docs.dyalog.com/20.0/language-reference-guide/system-commands/load/)

**Version of the documentation in which you are reporting the error**
2026-01-14 main:f8a6efd25f

**Describe the error**
Pages such as those listed above mention another system command/function without linking to the mentioned page.

This is especially bad for the system commands that are synonyms or provide near identical effects

PR #710, opened 2026-02-04
Ensure all references system functions use hyperlinks

bear8642:
As #708 states, make sure all document references use hyperlinks

### DIFF
--- a/language-reference-guide/docs/system-commands/copy.md
+++ b/language-reference-guide/docs/system-commands/copy.md
@@ -137,8 +137,8 @@ C:\Users\Pete\Desktop\pete.dse saved Wed Oct 14 ...
 ```
 
 - You can copy a function or variable from any namespace under a saved [`⎕SE`](../system-functions/se.md), no matter what your current namespace is.
-- You can copy any namespace without GUI from a saved [`⎕SE`](../system-functions/se.md), no matter what you current namespace is.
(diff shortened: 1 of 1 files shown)


language-reference-guide/docs/system-commands/copy.md
```diff
@@ -136,9 +136,9 @@ C:\Users\Pete\Desktop\pete.dse saved Wed Oct 14 ...
 
 ```
 
-- You can copy a function or variable from any namespace under a saved `⎕SE`, no matter what your current namespace is.
-- You can copy any namespace without GUI from a saved `⎕SE`, no matter what you current namespace is.
-- You can copy any namespace with GUI from a saved `⎕SE` as long as your current namespace is the same as the parent namespace of the namespace you are trying to copy
+- You can copy a function or variable from any namespace under a saved [`⎕SE`](../system-functions/se.md), no matter what your current namespace is.
+- You can copy any namespace without GUI from a saved [`⎕SE`](../system-functions/se.md), no matter what you current namespace is.
+- You can copy any namespace with GUI from a saved [`⎕SE`](../system-functions/se.md) as long as your current namespace is the same as the parent namespace of the namespace you are trying to copy
```
FionaDyalog: Only the first occurence of this should be a link, remove it for subsequent occurrences.

PR #727, opened 2026-02-24
348 reduction evaluation order qualification

FionaDyalog:
Reference: #348

### DIFF
--- a/language-reference-guide/docs/primitive-operators/scan.md
+++ b/language-reference-guide/docs/primitive-operators/scan.md
@@ -75,7 +75,7 @@ ONE BOOK
 
 ## Evaluation Order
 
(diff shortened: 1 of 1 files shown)


language-reference-guide/docs/primitive-operators/scan.md
```diff
@@ -73,4 +73,24 @@ ONE BOOK
  
 ```
 
+## Evaluation Order
 
+In its initial implementation, Dyalog evaluated `+\` and `×\` as a single left-to-right pass (non-ISO standard) order because `+` and `×` are commutative and left-to-right evaluation was faster.
```
FionaDyalog:
```suggestion
In its initial implementation, Dyalog evaluated `+\` and `×\` as a single left-to-right pass (non-ISO standard) because `+` and `×` are commutative and left-to-right evaluation was faster. 
```

Issue #751, opened 2026-04-09
Description of ODE menu options include obsolete items


FionaDyalog:
Re https://docs.dyalog.com/20.0/windows-ui-guide/session-menubar/#the-options-menu

Two menu items have been removed from the ODE (see Mantis 9859) and should be removed from the documentation:

- Expose Gui Properties
- Expose Session Properties

This does not need to be included in the Dyalog v21.0 Release Notes as it is not new for Dyalog v21.0.

PR #760, opened 2026-04-21
Options removed from Object menu

FionaDyalog:
Reference: #751


dyaandys (approved):
The update re expose properties looks good, but some of the other menus in the product differ from what is documented.

Issue #618, opened 2025-10-30
Dyalog v21.0 Release Notes (new document)


FionaDyalog:
THIS ISSUE IS A WORK IN PROGRESS (ADDITIONAL CONTENT WILL BE ADDED)

Create new set of Release Notes for Dyalog v21.0. These should be modelled on the Dyalog v20.0 Release Notes.

General sections needed:
- [x] Introduction
- [x] System Requirements
- [x] Announcements
- [x] Deprecated Functionality
- [x] New/Enhanced Functionality
- [x] Minor Updates
- [x] Interoperability

Project-specific work to include:
- [x] 00223.3 Platform features phase II
- [x] 00396 Convert formatted datetimes to numeric (inverse of 1200⌶)
- [x] 00398 Bump minimum supported WS version to 12.1
- [x] 00399 Deprecated component files - requirements for 21.0
- [x] 00406.1 Improved thread scheduler - phase I
- [x] 00407 Completion of support for generics in .NET bridges phase I
- [x] 00422 Allow ⎕UCS to accept/create UTF-8 in type 83 arrays
- [x] 00425 .NET v10 support
- [x] 00429 Better support for IDE Classic Mode
- [x] 00436 Substitution parameters in config settings
- [x] M22287/G724 New ⎕CSV variant option ForceQuotes
- [x] M22788/G747 New Properties for a Printer object
- [x] M22805 MinButton,MaxButton,SysMenu,HelpButton properties of a Form are now ⎕WSable
- [x] G990 Add a note about fixes to Stencil
- [x] G1008 ⎕TC has never generated tabstop
- [x] G1017 session log files have the extension .dlfx
- [x] G1018 Interoperability mentions writing to small-span files
- [x] G1051 Tweaks to 13 I-beam at Dyalog v21.0
- [x] G1070 Information about Dark Mode
- [x] 1077 Added the APL335 font

Misc To Do:
- [x] Create .yml file
- [x] Amend project .yml file (incl moving Dyalog v20.0 Release Notes into "Earlier Releases")
- [x] Update links in Dyalog v20.0 Release Notes to link to Dyalog v20.0 documentation (also needed as the links will be broken following the change in ToC hierarchy)
- [x] Ensure cross-document links are version-specific

FionaDyalog:
First tranche of work complete – moved Dyalog v20.0 Release Notes (with amended links) and created stubs for Dyalog v21.0 Release Notes.

PR #771, opened 2026-04-28
618 dyalog v21.0 release notes

FionaDyalog:
Reference: #618 

First tranche of work – setting up placeholder pages, moved v20.0 Release Notes, experimenting with version-specific links within Release notes.


release-notes/docs/index.md
```diff
@@ -0,0 +1,48 @@
...
+This document describes the requirements for installing Dyalog v21.0 and the functionality changes in Dyalog v21.0 compared with Dyalog v20.0.
+
+!!! Info "Information"  
+    If you are upgrading from Dyalog v19.0 or earlier to Dyalog v21.0, you are advised to read the Release Notes for any intermediate versions in conjunction with this document.
+	
+## Release Highlights
+
+Key enhancements in Dyalog v21.0 include the following:
+ 
+<!--
```
xpqz: Is there a reason to leave swathes of html comments in the markdown documents? Whilst technically legal, it seems like it's something bound to confuse both humans and tooling.

mkdocs.yml
```diff
@@ -71,7 +71,7 @@ markdown_extensions:
 
 nav:
   - 'Release Notes':
-        - 'v20.0 Release Notes': "!include ./release-notes/mkdocs.yml"
+        - 'v21.0 Release Notes': "!include ./release-notes/mkdocs.yml"
```
xpqz, line 74: A wider problem, perhaps, but we're now getting a mixture of quoting styles, singles, doubles... Note that the key side in yaml (before a colon) only needs quoting if the key itself contains a colon.

xpqz (changes requested):
- Avoid HTML comments. It basically makes it impossible to review, as the diff lights up in green for areas that are unchanged. In fact, avoid HTML in Markdown if at all possible (it's not always possible).
- Stick to one quote style in the yaml files -- note that keys need not be quoted, generally.

release-notes/docs/index.md
FionaDyalog: Purely for ease of adding new information to the document as it'll be based on the previous one – by the time it is published there won't be any "hidden" text in there as it will all either be repurposed or deleted.

xpqz (approved):
Approved following conversation on teams.

Issue #773, opened 2026-04-30
⎕ATX row is missing from the System Functions (by Category) page


bear8642:
**Page on which the error exists**
[System Functions (by Category)](https://docs.dyalog.com/20.0/language-reference-guide/system-functions/system-functions-by-category/)
Core Reference > Dyalog APL Language >System Functions

**Version of the documentation in which you are reporting the error**
main:1c22ea63b6

**Describe the error**
None of the various tables contain a ⎕ATX row

**Additional information**
We believe that ⎕ATX row below should be added to the Workspace table
Name | Description | Form
-- | -- | --
| ⎕ATX |  Extended Attributes |  Function |

PR #787, opened 2026-05-08
Add missing ⎕ATX row to the System Functions (by Category) page

bear8642:
As #773 states, `⎕ATX` is missing from the `System Functions (by Category) page". Add it


FionaDyalog (changes requested):
That change looks good but please also add to the main table on /language-reference-guide/system-functions/ (alphabetical order rather than categorised for that one).

bear8642:
Done. I can merge those into v20.0 too - are there any other pages that might need changing?

FionaDyalog (approved):
LGTM

PR #801, opened 2026-05-13
Rewrite json.md

### LINKED ISSUE #788: Document ⎕JSON wrappers
**Page on which the enhancement should be made**
https://docs.dyalog.com/20.0/language-reference-guide/system-functions/json/

**Describe the enhancement you would like**
> Codes 2, 3 and 4 are used to identify different forms of datasets.

Describe what they are.

Also fix the documentation for the JS injection which has escaped doublequotes.

abrudz:


DyalogRichard:
I have not been asked to review this, but as it's ⎕JSON I think I should. Whilst I was surprised that the page appears to have been rewritten so extensively to address #788, it is definitely the case that this page needed a revamp.

I had a *very* quick scan of the (GitHub-rendered) new page whilst travelling yesterday and there were a couple of technical things that I thought should be discussed before we publish - these may well have been inherited from the original page. However, it will take me a while before I can look at it more carefully and comment. I also had some much more general observations about tone and content which I think the docs team as a whole should review, particularly wrt the amount of specific detail it gives - for example, it now tries to be *very* precise about the effect of the Dialect variant option which Pete did not and would not do - he always seemed to give "just enough information" to make it clear what was happening, without getting bogged down, and in that context, saying it switches between JSON and JSON5 dialect support is all that is needed. JSON and JSON5 are well documented elsewhere and our documentation does not describe either; we describe just the system function itself. Given that we don't describe either JSON or JSON5, it seems to me that describing (just) the small subset of things that are different between the two is an odd thing to do. Similarly, the documentation for the Null variant option needs only say that it specifies whether the JSON null is represented by ⎕NULL or ⊂'null' - the table showing the four combinations of variant setting and import/exports is not strictly necessary - and I reckon it confuses rather than clarifies anyway: the cell for '⎕NULL' and export says "Both ⎕NULL and ⊂null (sic) are exported as JSON null", and whilst this is true, it is also misleading: ⎕NULL is exported as null because that APL type is represented that way, but ⊂'null' is exported as null because it now falls into the domain of Raw Text.

abrudz:
> I was surprised that the page appears to have been rewritten so extensively to address https://github.com/Dyalog/documentation/issues/788

I found multiple other issues, including inconsistent notes, examples disconnected from their subject matter description, inconsistent and misleading header structures, typos, bad phrasings, inconsistent usage of tables to describe variant behaviour, inconsistent adherence to best practices, unnecessarily complex APL constructs, and more.

> it now tries to be very precise about the effect of the Dialect variant (…) saying it switches between JSON and JSON5 dialect support is all that is needed

For import, yes (and only 10 words are spent on this), but that doesn't say anything about export, since JSON is valid JSON5.

> Given that we don't describe either JSON or JSON5, it seems to me that describing (just) the small subset of things that are different between the two is an odd thing to do.

It doesn't. It only specifies which features of JSON5 we apply on export: For example, when exporting compact JSON5, we could conceivably use hexadecimal representation of integers ≥10¹² and omit leading zeros for numbers in the range (−1,1) to obtain the most compact JSON5, but we don't. We could also prefer single quotes (many JS programmers do), but we don't.

> the documentation for the Null variant option (…) confuses rather than clarifies (…) and whilst this is true, it is also misleading

I agree, and have edited to remove confusing and misleading parts, instead adding a note with reference to the Raw Text section.

pmikkelsen (commented):
It looks good to me, although a few comments/questions:

* There is the possibility of turning many words into links, such as Null and Format. In some cases they are links to the relevant section describing the variant option, but not always. This is probably something I think we would benefit from having in the style guide as well.

* The variant options table would in my opinion be even more useful if there was a clear indication of which value is the default, for each option, if that makes sense.

* The "selection of a subset" section at the very end doesn't mention that the records and fields must be selected in increasing order (I found this by experimenting, so whether it is expected behaviour or a bug, I don't know). For example `1 ⎕JSON⊂4(invertedTable header)(1 2)` works, selecting records 1 and 2. But `1 ⎕JSON⊂4(invertedTable header)(1 1)` and `1 ⎕JSON⊂4(invertedTable header)(2 1)` does not work, even though from the documentation, I would expect it to include record 1 twice, and to include records 2 and 1, in that order.

FionaDyalog:
Re Peter's comment about indicating the default in the Variants table – please do not do this with shading but instead state it explicitly (replacing the obsolete "shading" convention in the documentation is an ongoing process, and it would be good not to introduce it here!).

language-reference-guide/docs/system-functions/json.md
```diff
@@ -9,96 +9,165 @@ search:
 
 <h1 class="heading"><span class="name">JSON Convert</span> <span class="command">R←{X}⎕JSON Y</span></h1>
 
-This function imports and exports data in [JavaScript Object Notation](https://www.json.org/json-en.html) (JSON) Data Interchange Format.
+!!! Warning "Warning"
+    The default for `X` depends on the type of `Y`. Dyalog Ltd strongly recommends that `X` should always be specified to avoid code that seemingly works, only to fail on specific values.
+
+This function imports and exports data in [JavaScript Object Notation](https://www.json.org/json-en.html) (JSON) data interchange format.
```
DyalogRichard: This is not needed - it is stated almost immediately below, and leading with it like this is just alarmist.

language-reference-guide/docs/system-functions/json.md
```diff
@@ -9,96 +9,165 @@ search:
...
+If specified, `X` must be a numeric scalar with the value `0` (import JSON) or `1` (export JSON). Dyalog Ltd strongly recommends that `X` should always be specified, however, if `X` is not specified and `Y` is a character array, `X` is assumed to be `0` (import); otherwise it is assumed to be `1` (export).
+
+!!! Hint "Hints and Recommendations"
+    As a mnemonic, think of `X` as specifying the desired "JSON-ness": `0` means "no JSON", that is, converting away from JSON; `1` means "yes JSON", that is, converting towards JSON.
+
+`⎕JSON` has six [variant options](#variant-options): `Format`, `Compact`, `Null`, `HighRank`, `Charset`, and `Dialect`, specified using [`⍠`](../primitive-operators/variant.md). The principal option is `Format`.
+
+## JSON Import
+
+If `X` is `0`, the JSON document `Y` is converted to a corresponding APL array or namespace `R`.
```
DyalogRichard: Is "document" the right term here? We use that term for the argument to ⎕R and ⎕S, but that is not the same.

Is "array or namespace" correct here? Elsewhere we refer to this as "data", and the different terms is confusing.  At least use "data" in the sentence too - e.g. "array or namespace (data)"

language-reference-guide/docs/system-functions/json.md
```diff
@@ -9,96 +9,165 @@ search:
...
+      0 ⎕JSON'[1,2,3]'
+1 2 3
 ```
 
-### Import as Data (`Format 'D'`)
+For details and more examples, see [Import to Data](#import-to-data) and [Import to Matrix](#import-to-matrix).
+
+## JSON Export
+
+If `X` is `1`, the APL array or namespace `Y` is converted to a corresponding JSON document `R`.
```
DyalogRichard: Ditto "document" and "data".

language-reference-guide/docs/system-functions/json.md
```diff
@@ -9,96 +9,165 @@ search:
...
-      j←0 ⎕JSON JSON
+      1 ⎕JSON 1 2 3
+[1,2,3]
+```
+
+For details and more examples, see [Export from Data](#export-from-data) and [Export from Matrix](#export-from-matrix).
+
+## Name Mangling
+
+When Dyalog imports from JSON to APL data, and a member of a JSON object has a name which is not a valid APL name, the member is renamed using a name mangling algorithm, resulting in a name that begins with `⍙`. Any characters that cannot be part of an APL name are replaced with their decimal Unicode code point surrounded by `⍙`s.
```
DyalogRichard: The previous text just had "resulting in a name that begins with ⍙". The addition is pretty much self-evident and seems like an unnecessary distraction.

language-reference-guide/docs/system-functions/json.md
```diff
@@ -9,96 +9,165 @@ search:
...
+| [**`HighRank`**](#variant-option-highrank) | `'Error'`<br><small>default</small>   | None                                                  | High-rank arrays are rejected |
+|_-                                        -_| `'Split'`                             |_-                                                   -_| High-rank arrays are split and [inverted table wrappers](#dataset-wrappers) accept text columns as matrices |
+
+### Variant Option: Format
+
+The `Format` variant option, the principal option, determines whether `⎕JSON` works with a direct APL representation of the data (`'D'` for "Data", the default) or with a four-column matrix that encodes the JSON structure (`'M'` for "Matrix") as nodes with depth, name, value, and type.
+
+#### Import to Data
+
+If `Format` is `'D'` (which stands for "Data", the default) the JSON document in `Y` is converted to the corresponding APL data `R` which is an array or a namespace, possibly containing arrays and/or sub-namespaces.
```
DyalogRichard: Repeats meaning of 'D'

DyalogRichard: Why not sub-arrays?

'[1, [2, 3]]'

has got sub-arrays.

language-reference-guide/docs/system-functions/json.md
```diff
@@ -275,297 +395,387 @@ If there are any mis-matches between the values in `Y[;3]` and the types in `Y[;
...
-      ns←⎕NS ''
-      ns.dé←'DÉ'
-      ns.dé
-DÉ
-      ⎕JSON ns
-{"dé":"DÉ"}
-      (⎕JSON⍠'Charset' 'ASCII')ns
-{"d\u00E9":"D\u00C9"}
-```
+If the `Dialect` variant option (default: `'JSON'`) is `'JSON5'`, [JSON5](https://json5.org/) extensions are enabled on import and export: Comments, hexadecimal literals, leading/trailing decimal points, character escapes of the form `\xNN`, trailing commas, unquoted ECMAScript 5.1 identifiers as object member names, single-quoted strings, and `Infinity`/`-Infinity`/`NaN` are accepted and/or produced.
```
DyalogRichard: I do not think we should enumerate the differences between JSON and JSON5 - this is not a JSON primer, and we leave definition of the syntax of JSON and JSON5 to their official docs. And this list is incomplete (off the top of my head: numbers can begin with 0; there are more supported single character escape sequences in strings, for example).

DyalogRichard:
This is way better than what went before. Some comments to argue about :-)

language-reference-guide/docs/system-functions/json.md
```diff
@@ -9,98 +9,196 @@ search:
 
 <h1 class="heading"><span class="name">JSON Convert</span> <span class="command">R←{X}⎕JSON Y</span></h1>
 
-This function imports and exports data in [JavaScript Object Notation](https://www.json.org/json-en.html) (JSON) Data Interchange Format.
+This function imports and exports data in [JavaScript Object Notation](https://www.json.org/json-en.html) (JSON) data interchange format.
+
+<h3 class="example">Examples</h3>
```
FionaDyalog, line 18: I don't think this should be here; I suggest moving it into the "Syntax" section that follows either before or after the Hint admonition. Alternatively, delete it – these examples are included in the "JSON Import" and "JSON Export" sections that follow soon after this.

language-reference-guide/docs/system-functions/json.md
```diff
@@ -9,98 +9,196 @@ search:
...
+
+!!! Hint "Hints and Recommendations"
+    As a mnemonic, think of `X` as specifying the desired "JSON-ness": `0` means "no JSON", that is, converting away from JSON; `1` means "yes JSON", that is, converting towards JSON.
+
+If `X` is not specified (that is, `⎕JSON` is called monadically), its assumed value depends on `Y`:  If `Y` is a character array, `X` is assumed to be `0`; otherwise it is assumed to be `1`.
+
+!!! Warning "Warning"
+    Dyalog Ltd strongly recommends that `X` should always be specified to avoid code that seemingly works, only to fail on specific values.
+
+`⎕JSON` has six [variant options](#variant-options): `Format`, `Compact`, `Null`, `HighRank`, `Charset`, and `Dialect`, specified using [`⍠`](../primitive-operators/variant.md). The principal option is `Format`.
```
FionaDyalog: I assume that adding the six individual sub-section links to the individual variant options looks like link overload, hence why you have not done it!

language-reference-guide/docs/system-functions/json.md
```diff
@@ -9,98 +9,196 @@ search:
...
+## Limitations
 
 JSON supports a limited number of data types and there is not a direct correspondence between JSON and APL data structures. In particular:
 
-- JSON does not support arrays with rank &gt;1.
-- JSON standard includes Boolean values true and false which are distinct from numeric values 1 and 0, and have no direct APL equivalent.
-- The [JSON5](https://json5.org/) standard includes numeric constants `Infinity`, `-Infinity`, `NaN` and `-NaN` which have no direct APL equivalent.
+- JSON does not support arrays with rank &gt; 1.
+- JSON does not support nested scalars.
+- JSON includes Boolean values `true` and `false` which are distinct from numeric values `1` and `0` and have no direct APL equivalent.
```
FionaDyalog:
```suggestion
- JSON includes Boolean values <code class="language-nonAPL">true</code> and <code class="language-nonAPL">false</code> which are distinct from numeric values <code class="language-nonAPL">1</code> and <code class="language-nonAPL">0</code>; these have no direct APL equivalent.
```

language-reference-guide/docs/system-functions/json.md
```diff
@@ -9,98 +9,196 @@ search:
...
+If `X` is not specified (that is, `⎕JSON` is called monadically), its assumed value depends on `Y`:  If `Y` is a character array, `X` is assumed to be `0`; otherwise it is assumed to be `1`.
+
+!!! Warning "Warning"
+    Dyalog Ltd strongly recommends that `X` should always be specified to avoid code that seemingly works, only to fail on specific values.
+
+`⎕JSON` has six [variant options](#variant-options): `Format`, `Compact`, `Null`, `HighRank`, `Charset`, and `Dialect`, specified using [`⍠`](../primitive-operators/variant.md). The principal option is `Format`.
+
+## Limitations
 
 JSON supports a limited number of data types and there is not a direct correspondence between JSON and APL data structures. In particular:
```
FionaDyalog:
```suggestion
JSON supports a limited number of data types, and there is no direct correspondence between JSON and APL data structures. In particular:
```

language-reference-guide/docs/system-functions/json.md
```diff
@@ -9,98 +9,196 @@ search:
...
 
-This function imports and exports data in [JavaScript Object Notation](https://www.json.org/json-en.html) (JSON) Data Interchange Format.
+This function imports and exports data in [JavaScript Object Notation](https://www.json.org/json-en.html) (JSON) data interchange format.
+
+<h3 class="example">Examples</h3>
+
+Importing a JSON document to APL:
+```apl
+      0 ⎕JSON'[1,-2,3]'
+1 ¯2 3
```
FionaDyalog, line 23:
```suggestion
- JSON object members are named, and these names might not be [valid names in APL](../../programming-reference-guide/introduction/names/).
```

language-reference-guide/docs/system-functions/json.md
```diff
@@ -9,98 +9,196 @@ search:
...
 
-- JSON does not support arrays with rank &gt;1.
-- JSON standard includes Boolean values true and false which are distinct from numeric values 1 and 0, and have no direct APL equivalent.
-- The [JSON5](https://json5.org/) standard includes numeric constants `Infinity`, `-Infinity`, `NaN` and `-NaN` which have no direct APL equivalent.
+- JSON does not support arrays with rank &gt; 1.
+- JSON does not support nested scalars.
+- JSON includes Boolean values `true` and `false` which are distinct from numeric values `1` and `0` and have no direct APL equivalent.
 - JSON object members are named and these names might not be valid names in APL.
-                
+- The [JSON5](https://json5.org/) dialect includes numeric constants `Infinity`, `-Infinity`, and `NaN`, which have no direct APL equivalent.
```
FionaDyalog: The [JSON5](https://json5.org/) dialect includes numeric constants <code class="language-nonAPL">Infinity</code>, <code class="language-nonAPL">-Infinity</code>, and <code class="language-nonAPL">NaN</code>, which have no direct APL equivalent.

language-reference-guide/docs/system-functions/json.md
```diff
@@ -9,98 +9,196 @@ search:
...
-    null       
-   ]            
-  }              
- }                
-}
+## JSON Import
+
+If `X` is `0`, the JSON document `Y` is converted to the corresponding APL data `R`.
+
+`Y` is a character scalar, vector, or matrix in JSON format. There is an implied newline character between each row of a matrix. By default, `R` is APL data, possibly containing sub-arrays and/or sub-namespaces. With [`Format`](#variant-option-format) being `'M'`, `R` is instead a matrix that represents the JSON structure.
```
FionaDyalog:
```suggestion
`Y` is a character scalar, vector, or matrix in JSON format. There is an implied newline character between each row of a matrix.

By default, `R` is APL data, possibly containing sub-arrays and/or sub-namespaces. If the variant option [`Format`](#variant-option-format) is set to `'M'`, `R` is instead a matrix that represents the JSON structure.
```

language-reference-guide/docs/system-functions/json.md
```diff
@@ -9,98 +9,196 @@ search:
...
+
+The [JSON standard says](https://www.rfc-editor.org/info/rfc8259/#section-4) that members of a JSON object should have unique names and that implementations vary in how they treat duplicates. Dyalog does not error on duplicates, but their handling depends on the `Format` variant.
```
FionaDyalog:
```suggestion
The [JSON standard](https://www.rfc-editor.org/info/rfc8259/#section-4) states that members of a JSON object should have unique names and that implementations vary in how they treat duplicates. Dyalog does not error on duplicates, but their handling depends on the `Format` variant option.
```

language-reference-guide/docs/system-functions/json.md
```diff
@@ -9,98 +9,196 @@ search:
 
 <h1 class="heading"><span class="name">JSON Convert</span> <span class="command">R←{X}⎕JSON Y</span></h1>
```
FionaDyalog, line 11: It's not ideal, but throughout the rest of this document variant option are presented in bold text and not in APL font. They should all be in APL font, and I will raise an issue about this. However, to be consistent with the rest of the documentation this should follow the same style as the rest of the document set and then they can all be amended as the same time.

I have not made this change in any of my individual comments – my hope is that we will have the capacity to update all the other references rather than change these ones onyl to need to change them back again later!

language-reference-guide/docs/system-functions/json.md
```diff
@@ -9,98 +9,196 @@ search:
...
+1 ¯2 3
+```
+
+For details and more examples, see [Import to Data](#import-to-data) and [Import to Matrix](#import-to-matrix).
+
+## JSON Export
+
+If `X` is `1`, the APL data `Y` is converted to a corresponding JSON document `R`.
+
+`Y` is the data to be exported. By default, `Y` must be APL data that can be represented as JSON (subject to the [`HighRank`](#variant-option-highrank) option). With [`Format`](#variant-option-format) being `'M'`, `Y` must instead be a matrix representation such as would have been produced by importing JSON with `Format` being `'M'`. `⎕JSON` will signal `DOMAIN ERROR` if `Y` is incompatible with the specified (or implied) value of `Format`.
```
FionaDyalog:
```suggestion
`Y` is the data to be exported. By default, `Y` must be APL data that can be represented as JSON (subject to the [`HighRank`](#variant-option-highrank) variant option). If the [`Format`](#variant-option-format) variant option is set to `'M'`, `Y` must instead be a matrix representation such as would have been produced by importing JSON with `Format` being `'M'`. `⎕JSON` will signal `DOMAIN ERROR` if `Y` is incompatible with the specified (or implied) value of `Format`.
```

language-reference-guide/docs/system-functions/json.md
```diff
@@ -9,98 +9,196 @@ search:
...
+
+`R` is a character vector whose content depends upon the values of the [`Compact`](#variant-option-compact), [`Dialect`](#variant-option-dialect), and [`Charset`](#variant-option-charset) variants.
```
FionaDyalog:
```suggestion
`R` is a character vector whose content depends upon the values of the [`Compact`](#variant-option-compact), [`Dialect`](#variant-option-dialect), and [`Charset`](#variant-option-charset) variant options.
```

language-reference-guide/docs/system-functions/json.md
```diff
@@ -9,98 +9,196 @@ search:
...
+
+Some JSON values lack a direct APL equivalent (`true`, `false`, `null`, JavaScript fragments), and some APL representations of datasets do not map directly to JSON. Such cases are handled by [wrappers](#wrappers).
```
FionaDyalog:
```suggestion
Some JSON values lack a direct APL equivalent (<code class="language-nonAPL">true</code>, <code class="language-nonAPL">false</code>, <code class="language-nonAPL">null</code>, JavaScript fragments), and some APL representations of datasets do not map directly to JSON. Such cases are handled by [wrappers](#wrappers).
```

language-reference-guide/docs/system-functions/json.md
```diff
@@ -9,98 +9,196 @@ search:
...
+```apl
+      1 ⎕JSON 1 ¯2 3
+[1,-2,3]
+```
+
+For details and more examples, see [Export from Data](#export-from-data) and [Export from Matrix](#export-from-matrix).
+
+## Name Mangling
+
+When `⎕JSON` converts a JSON document to APL data, and a member of a JSON object has a name which is not a valid APL name, the member is renamed using a name mangling algorithm, resulting in a name that begins with `⍙`. Any characters that cannot be part of an APL name are replaced with their [decimal Unicode code point](ucs.md#monadic-ucs) surrounded by `⍙`s.
```
FionaDyalog:
```suggestion
When `⎕JSON` converts a JSON document to APL data and a member of a JSON object has a name that is not a valid APL name, the member is renamed using a name mangling algorithm. This results in a name that begins with `⍙`. Any characters that cannot be part of an APL name are replaced with their [decimal Unicode code point](ucs.md#monadic-ucs) surrounded by `⍙`s.
```

language-reference-guide/docs/system-functions/json.md
```diff
@@ -9,98 +9,196 @@ search:
...
+
+[`7162⌶`](../primitive-operators/i-beam/json-translate-name.md) provides direct access to the name mangling algorithm.
+
+<h3 class="example">Examples</h3>
```
FionaDyalog: The distinction between the two examples is not obvious. Maybe go with two different example headings with a few explanatory words, for example:

Example: Object member includes non-valid character

language-reference-guide/docs/system-functions/json.md
```diff
@@ -9,98 +9,196 @@ search:
...
+      1 ⎕JSON (a:1 ⋄ ⍙2a:2)
+{"a":1,"2a":2}
 ```
 
-### Import as Data (`Format 'D'`)
+This object has a member name with a character (`ý`; `⎕UCS 253`) that is not allowed in APL names:
+```json
+{"sýn":"vision"}
+```
+The `ý` is replaced with `⍙253⍙`:
```
FionaDyalog:
```suggestion
The `ý` is replaced with `⍙253⍙` ("253" is the Unicode decimal character code for this character):
```

language-reference-guide/docs/system-functions/json.md
```diff
@@ -9,98 +9,196 @@ search:
...
+└───┴────────┘
+      1(7162⌶)'⍙2a' '⍙s⍙253⍙n'
+┌──┬───┐
+│2a│sýn│
+└──┴───┘
+```
+
+## Variant Options
+
+`⎕JSON` is controlled by six variant options. The following table summarises each option's effect on import from JSON to APL (`X=0`) and export from APL to JSON (`X=1`). Each option is described in finer detail, with examples, below the table. Variant options specific to one direction are tolerated for the other direction even if they have no effect.
```
FionaDyalog:
```suggestion
`⎕JSON` is controlled by six variant options. [](#variant-table) summarises each option's effect on import from JSON to APL (`X=0`) and export from APL to JSON (`X=1`). Each option is described in finer detail, with examples, below the table. Variant options specific to one direction are tolerated for the other direction even if they have no effect.
```

language-reference-guide/docs/system-functions/json.md
```diff
@@ -9,98 +9,196 @@ search:
...
+| [**`Null`**](#variant-option-null)                                 | `⊂'null'`<br><small>default</small>   | JSON `null` becomes APL `⊂'null'`                   | APL `⊂'null'` becomes JSON `null` |
+|_-                                                                -_| `⎕NULL`                               | JSON `null` becomes APL `⎕NULL`                     | APL `⎕NULL` becomes JSON `null` |
+| [**`Compact`**](#variant-option-compact)                           | `1`<br><small>default</small>         | None                                                | `R` has no whitespace outside quotes |
+|_-                                                                -_| `0`                                   | None                                                | `R` has whitespace for readability and, if `Dialect` is `'JSON5'`, trailing commas after final elements and members |
+| [**`Charset`**](#variant-option-charset)                           | `'Unicode'`<br><small>default</small> | None                                                | Unicode characters in `Y` are included verbatim when JSON standard allows |
```
FionaDyalog: replace verbatim with a non-Latin word...exactly/as written/precisely (or the English alternative of your choice!)

language-reference-guide/docs/system-functions/json.md
```diff
@@ -9,98 +9,196 @@ search:
...
+| [**`HighRank`**](#variant-option-highrank)                         | `'Error'`<br><small>default</small>   | None                                                | High-rank arrays are rejected |
+|_-                                                                -_| `'Split'`                             | None                                                | High-rank arrays are split and [inverted table wrappers](#dataset-wrappers) accept text columns as matrices |
+
+### Variant Option: Format
+
+The `Format` variant option, the principal option, determines whether `⎕JSON` works with a direct APL representation of the data (`'D'` for "Data", the default) or with a four-column matrix that encodes the JSON structure (`'M'` for "Matrix") as nodes with depth, name, value, and type.
+
+#### Import to Data
 
+If `Format` is `'D'` (which stands for "Data", the default) the JSON document in `Y` is converted to the corresponding APL data `R`, possibly containing sub-arrays and/or sub-namespaces.
```
FionaDyalog:
```suggestion
If `Format` is `'D'` (which stands for "Data", the default), the JSON document in `Y` is converted to the corresponding APL data `R`, possibly containing sub-arrays and/or sub-namespaces:
```

language-reference-guide/docs/system-functions/json.md
```diff
@@ -9,98 +9,196 @@ search:
...
+| [**`Format`**](#variant-option-format)<br><small>principal</small> | `'D'`<br><small>default</small>       | `R` is APL data corresponding to `Y`                | `Y` is APL data |
+|_-                                                                -_| `'M'`                                 | `R` is an APL matrix encoding of `Y`                | `Y` is a 4-column APL matrix as from import with `'M'` |
+| [**`Dialect`**](#variant-option-dialect)                           | `'JSON'`<br><small>default</small>    | Only strict JSON syntax is accepted                 | Only strict JSON syntax is produced |
+|_-                                                                -_| `'JSON5'`                             | [JSON5](https://json5.org/) extensions are accepted | JSON5 features are used to improve readability and editability, and/or shorten output |
+| [**`Null`**](#variant-option-null)                                 | `⊂'null'`<br><small>default</small>   | JSON `null` becomes APL `⊂'null'`                   | APL `⊂'null'` becomes JSON `null` |
```
FionaDyalog:
```suggestion
| [**`Null`**](#variant-option-null)                                 | `⊂'null'`<br><small>default</small>   | JSON <code class="language-nonAPL">null</code> becomes APL `⊂'null'`                   | APL `⊂'null'` becomes JSON <code class="language-nonAPL">null</code> |
```

language-reference-guide/docs/system-functions/json.md
```diff
@@ -9,98 +9,196 @@ search:
...
+|_-                                                                -_| `⎕NULL`                               | JSON `null` becomes APL `⎕NULL`                     | APL `⎕NULL` becomes JSON `null` |
```
FionaDyalog:
```suggestion
|_-                                                                -_| `⎕NULL`                               | JSON <code class="language-nonAPL">null</code> becomes APL `⎕NULL`                     | APL `⎕NULL` becomes JSON <code class="language-nonAPL">null</code> |
```

language-reference-guide/docs/system-functions/json.md
```diff
@@ -9,98 +9,196 @@ search:
...
+
+The `Format` variant option, the principal option, determines whether `⎕JSON` works with a direct APL representation of the data (`'D'` for "Data", the default) or with a four-column matrix that encodes the JSON structure (`'M'` for "Matrix") as nodes with depth, name, value, and type.
+
+#### Import to Data
 
+If `Format` is `'D'` (which stands for "Data", the default) the JSON document in `Y` is converted to the corresponding APL data `R`, possibly containing sub-arrays and/or sub-namespaces.
+
+- JSON arrays are converted into APL vectors.
+- JSON objects are converted into APL namespaces.
+- JSON `null` is converted into the specified (or implied) value of [**`Null`**](#variant-option-null) (`⊂'null'`, the default, or `⎕NULL`).
```
FionaDyalog:
```suggestion
- JSON <code class="language-nonAPL">null</code> is converted into the specified (or implied) value of [**`Null`**](#variant-option-null) (`⊂'null'`, the default, or `⎕NULL`).
```

language-reference-guide/docs/system-functions/json.md
```diff
@@ -9,98 +9,196 @@ search:
...
+- JSON `true` and `false` and, if the [**`Dialect`**](#variant-option-dialect) variant option is `'JSON5'`, the JSON5 numeric constants `Infinity`, `-Infinity`, and `NaN`, are converted to enclosed character vectors `⊂'true'`, `⊂'false'`, and so forth.
```
FionaDyalog:
```suggestion
- JSON <code class="language-nonAPL">true</code> and <code class="language-nonAPL">false</code> and, if the [**`Dialect`**](#variant-option-dialect) variant option is `'JSON5'`, the JSON5 numeric constants <code class="language-nonAPL">Infinity</code>, <code class="language-nonAPL">-Infinity</code>, and <code class="language-nonAPL">NaN</code>, are converted to enclosed character vectors `⊂'true'`, `⊂'false'`, and so on.
```

language-reference-guide/docs/system-functions/json.md
```diff
@@ -9,98 +9,196 @@ search:
...
+- If the JSON source contains object member names which are not valid APL names they are converted to APL namespace members with [mangled names](#name-mangling). The original names can be obtained using [`7162⌶`](../primitive-operators/i-beam/json-translate-name.md).
```
FionaDyalog:
```suggestion
- If the JSON source contains object member names that are not valid APL names, they are converted to APL namespace members with [mangled names](#name-mangling). The original names can be obtained using [`7162⌶`](../primitive-operators/i-beam/json-translate-name.md).
```

language-reference-guide/docs/system-functions/json.md
```diff
@@ -9,98 +9,196 @@ search:
...
+- If duplicate names are found, the last member encountered is used and all previous members with the same name are discarded.
+
+<h5 class="example">Examples</h5>
+
+This JSON document (stored as the character vector `json`) is used in the following example:
```
FionaDyalog:
```suggestion
The following JSON document is stored as the character vector `json`:
```

language-reference-guide/docs/system-functions/json.md
```diff
@@ -116,25 +214,87 @@ c
...
+| `6`     | Enclosed character vector              | Lacking APL equivalent   |
+
+Note that:
+
+- JSON `null` is converted into the specified (or implied) value of [**`Null`**](#variant-option-null); `⊂'null'` (the default) or `⎕NULL`.
+- JSON values that lack an APL equivalent, `true` and `false`, and, if `Dialect` is `'JSON5'`, the JSON5 numeric constants `Infinity`, `-Infinity`, and `NaN`, are converted to enclosed character vectors `⊂'true'`, `⊂'false'`, and so forth.
+- Object member names are reported as specified in the JSON text; they are not mangled as when `Format` is `'D'`.
+- If duplicate names are found, all duplicate members are recorded in the result matrix.
+
+<h5 class="example">Example</h5>
```
FionaDyalog, line 285:
```suggestion
<h5 class="example">Example</h5>
This example uses the character vector `json` from the previous example:
```

language-reference-guide/docs/system-functions/json.md
```diff
@@ -116,25 +214,87 @@ c
...
+| `1`     | Empty (contents are in following rows) | Object                   |
+| `2`     | Empty (contents are in following rows) | Array                    |
+| `3`     | Number                                 | Number                   |
+| `4`     | Character vector                       | String                   |
+| `5`     | Specified by `Null` variant            | Null                     |
+| `6`     | Enclosed character vector              | Lacking APL equivalent   |
+
+Note that:
+
+- JSON `null` is converted into the specified (or implied) value of [**`Null`**](#variant-option-null); `⊂'null'` (the default) or `⎕NULL`.
```
FionaDyalog:
```suggestion
- JSON <code class="language-nonAPL">null</code> is converted into the specified (or implied) value of [**`Null`**](#variant-option-null); `⊂'null'` (the default) or `⎕NULL`.
```

language-reference-guide/docs/system-functions/json.md
```diff
@@ -116,25 +214,87 @@ c
...
+- JSON values that lack an APL equivalent, `true` and `false`, and, if `Dialect` is `'JSON5'`, the JSON5 numeric constants `Infinity`, `-Infinity`, and `NaN`, are converted to enclosed character vectors `⊂'true'`, `⊂'false'`, and so forth.
```
FionaDyalog:
```suggestion
- JSON values that lack an APL equivalent, <code class="language-nonAPL">true</code> and <code class="language-nonAPL">false</code>, and, if `Dialect` is `'JSON5'`, the JSON5 numeric constants <code class="language-nonAPL">Infinity</code>, <code class="language-nonAPL">-Infinity</code>, and <code class="language-nonAPL">NaN</code>, are converted to enclosed character vectors `⊂'true'`, `⊂'false'`, and so on.
```

FionaDyalog (changes requested):
First tranche of review comments: Done to the start of Format > Export from Data

language-reference-guide/docs/system-functions/json.md
```diff
@@ -170,87 +330,77 @@ e
...
+|---------|---------------------------|--------------------------|
+| `1`     | Empty array               | Object                   |
+| `2`     | Empty array               | Array                    |
+| `3`     | Numeric scalar            | Number                   |
+| `4`     | Character vector          | String                   |
+| `5`     | Null                      | Null                     |
+| `6`     | Enclosed character vector | Lacking APL equivalent   |
+| `7`     | Enclosed character vector | [Raw text](#raw-text-wrapper) |
+
+The difference between JSON types `6` and `7` is that `6` only allows the special values that can be imported, while `7` allows any text whatsoever.
```
FionaDyalog:
```suggestion
The difference between JSON types `6` and `7` is that `7` allows any text but `6` only allows the special values that can be imported.
```

language-reference-guide/docs/system-functions/json.md
```diff
@@ -260,11 +410,11 @@ If there are any mis-matches between the values in `Y[;3]` and the types in `Y[;
 ├─┼──────┼───┼─┤
 │2│      │300│3│
 └─┴──────┴───┴─┘
-
-      M[3;3]←⊂'75' ⍝ character not numeric
-
-      M            ⍝ but looks the same as before
-
+```
+Modifying the above matrix, one number is replaced with an identically looking character vector to illustrate type mismatches:
```
FionaDyalog:
```suggestion
To illustrate type mismatches, modify the above matrix by replacing one number with a character vector that looks the same:
```

language-reference-guide/docs/system-functions/json.md
```diff
@@ -274,298 +424,395 @@ If there are any mis-matches between the values in `Y[;3]` and the types in `Y[;
...
+┌───┬───┬──┐
+│abc│0.1│15│
+└───┴───┴──┘
 ```
 
-### Raw Text
+### Variant Option: Null
 
-An enclosed character vector is inserted into the result of JSON export as raw text. This feature may be used to export special JSON values such as `null`, `true` and `false`. Without the extra enclosure, the character vectors are exported as strings:
+The `Null` variant option selects how JSON `null` is represented in APL, and must be either `⊂'null'` (the default) or `⎕NULL`. With `Null` being `⊂'null'`, `⎕NULL` causes `DOMAIN ERROR`. With `Null` being `⎕NULL`, `⊂'null'` is still exported as `null` because it is interpreted as [raw text](#raw-text-wrapper).
```
FionaDyalog:
```suggestion
The `Null` variant option selects how JSON <code class="language-nonAPL">null</code> is represented in APL, and must be either `⊂'null'` (the default) or `⎕NULL`:

- If `Null` is `⊂'null'`, `⎕NULL` causes `DOMAIN ERROR`.
- If `Null` is `⎕NULL`, `⊂'null'` is still exported as `null` because it is interpreted as [raw text](#raw-text-wrapper).
```

language-reference-guide/docs/system-functions/json.md
```diff
@@ -274,298 +424,395 @@ If there are any mis-matches between the values in `Y[;3]` and the types in `Y[;
...
-      fn2,←' " - $" + ui.values[ 1 ] );}'
+The `Compact` variant option can be used to generate JSON that is either dense (`1`, the default) or optimised for humans to read and edit (`0`).
 
-      Slider.slide←,/fn1 fn2 ⍝ Enclosed character vec
-      ⍴JS←1 ⎕JSON Slider
-159
-      JS
-{"max":500,"min":0,"range":true,"slide": function( event, ui ) {$( \"#amount\" ).val( \"$\" + ui.values[ 0 ] + \" - $\" + ui.values[ 1 ]);},"values":[75,300]}
-```
+With `Compact` being `0`:
```
FionaDyalog:
```suggestion
If `Compact` is `0`:
```

language-reference-guide/docs/system-functions/json.md
```diff
@@ -274,298 +424,395 @@ If there are any mis-matches between the values in `Y[;3]` and the types in `Y[;
...
-As an aside, note that using this scheme each variable represents an inverted index into the dataset and enables rapid searches.
+### Variant Option: HighRank
 
-```apl
-      (Price<4)/Items
-┌─────┐
-│Knife│
-└─────┘
-```
+If `HighRank` is `'Error'` (the default), `⎕JSON` will signal a `DOMAIN ERROR` upon encountering any arrays in `Y` of rank higher than 1. If `HighRank` is `'Split'`, `⎕JSON` will recursively split any such arrays as necessary. This also allows [datasets](#dataset-wrappers) as inverted tables with text columns as matrices.
```
FionaDyalog: Hard to parse last sentence (I can't suggest a rewrite without understanding it). Do you mean that Y can be an inverted table with text columns as matrices, that the variant option can perform such an inversion, that the result can be presented/transformed in that way, ... please redraft to clarify.

language-reference-guide/docs/system-functions/json.md
```diff
@@ -274,298 +424,395 @@ If there are any mis-matches between the values in `Y[;3]` and the types in `Y[;
...
 
-┌────────────────┬────────────┐
-│┌────┬─────┬───┐│┌─────┬─┬──┐│
-││Item│Price│Qty│││Knife│3│23││
-│└────┴─────┴───┘│├─────┼─┼──┤│
-│                ││Fork │4│45││
-│                │└─────┴─┴──┘│
-└────────────────┴────────────┘
-```
+This structure has been chosen to identify special handling because a nested scalar cannot be represented in JSON or JavaScript. A wrapper can be specified directly in the right argument to `⎕JSON` and/or as part of the data structure specified by the right argument, as a sub-array or in a sub-namespace. This allows a special array to be processed appropriately as part of a general data structure that is to be rendered as JSON.
```
FionaDyalog:
```suggestion
Some APL representations of datasets do not map directly to JSON or JavaScript, for example, nested scalars. A wrapper provides `special` handling if such an APL representation is encountered. It can be declared in the right argument to `⎕JSON` and/or as part of the data structure specified by the right argument, as a sub-array or in a sub-namespace. This allows a `special` array to be processed appropriately as part of a general data structure that is to be rendered as JSON.
```

language-reference-guide/docs/system-functions/json.md
```diff
@@ -274,298 +424,395 @@ If there are any mis-matches between the values in `Y[;3]` and the types in `Y[;
...
 
-A third way retains the inverted nature of the data structure, storing the values as a vector. The advantage of this structure is that it consumes significantly less memory compared to the matrix forms, because numeric columns are stored as simple numeric vectors.
+The structure of the `special` array is specified within the wrapper by a leading numeric code. Code `1` (the default) allows insertion of raw text, including JSON values such as `null` and `true`. Codes `2`, `3` and `4` identify various representations of a *dataset*.
```
FionaDyalog:
```suggestion
The structure of the `special` array is specified within the wrapper by a leading numeric code. Code `1` (the default) allows insertion of raw text, including JSON values such as <code class="language-nonAPL">null</code> and <code class="language-nonAPL">true</code>. Codes `2`, `3`, and `4` identify various representations of a *dataset*.
```

language-reference-guide/docs/system-functions/json.md
```diff
@@ -274,298 +424,395 @@ If there are any mis-matches between the values in `Y[;3]` and the types in `Y[;
...
 
-```apl
-      (Fields (Items Price Qty))
+The term *dataset* is used here to mean a collection of data, usually presented in tabular form. Each named column (also called a *field*) represents a particular variable. Each row (also called a *record*) corresponds to a given member of the dataset in question, listing its value for each of the variables, such as price and quantity of an item.
```
FionaDyalog: Add this to end of previous paragraph rather than have it as a separate paragraph.

language-reference-guide/docs/system-functions/json.md
```diff
@@ -274,298 +424,395 @@ If there are any mis-matches between the values in `Y[;3]` and the types in `Y[;
...
+  {
+    "item": "Spoon",
+    "price": 5,
+    "qty": 67
+  }
 ]
 ```
 
-**Examples**
+Note that the JSON structure can be represented in APL:
```
FionaDyalog:
```suggestion
The JSON structure can be represented in APL as:
```

language-reference-guide/docs/system-functions/json.md
```diff
@@ -274,298 +424,395 @@ If there are any mis-matches between the values in `Y[;3]` and the types in `Y[;
...
+If such a representation is already used in an APL application, then no special handling is necessary to generate the corresponding JSON. However, transforming a dataset into a vector of namespaces, just for export to JSON, is cumbersome and can be expensive. `⎕JSON`'s wrapper codes `2`, `3`, and `4`, provide a quick and efficient way to transform the common APL representations of a dataset directly into a JSON array of objects.
 
-```apl
-      ⎕JSON Fields,[1]↑[1]Items Price Qty)
+### Raw Text Wrapper
 
-DOMAIN ERROR: JSON export: the right argument cannot be converted (⎕IO=1)
-⎕JSON Fields,[1]↑[1]Items Price Qty)
-∧
+Special JSON values such as `null`, `true` and `false` do not directly correspond to specific APL values and therefore require special handling. This is provided by wrapper code `1`:
```
FionaDyalog:
```suggestion
Special JSON values such as <code class="language-nonAPL">null</code>, <code class="language-nonAPL">true</code> and <code class="language-nonAPL">false</code> do not directly correspond to specific APL values and, therefore, require special handling. This is provided by wrapper code `1`:
```

language-reference-guide/docs/system-functions/json.md
```diff
@@ -274,298 +424,395 @@ If there are any mis-matches between the values in `Y[;3]` and the types in `Y[;
...
+
+```apl
+      1 ⎕JSON 42 'text'(⊂1 'null')(⊂1 'true')(⊂1 'false')
+[42,"text",null,true,false]
+```
+The code number can be omitted:
```
FionaDyalog:
```suggestion
As `1` is the default code number, it can be omitted:
```

language-reference-guide/docs/system-functions/json.md
```diff
@@ -274,298 +424,395 @@ If there are any mis-matches between the values in `Y[;3]` and the types in `Y[;
...
+!!! Warning "Warning"
+    Out of convenience, it is common to initialise a list using a scalar rather than a one-element vector. Doing so for an eventual vector of character vectors will fail if the value is exported before a subsequent element is added:
+    ```apl
+          list←⊂'foo'
+          1 ⎕JSON list
+    foo
+          1 ⎕JSON list,'bar' 'baz'
+    ["foo","bar","baz"]
+    ```
+    It is therefore important to initialise as a proper vector:
```
FionaDyalog:
```suggestion
    It is, therefore, important to convert to a vector before initialising:
```

language-reference-guide/docs/system-functions/json.md
```diff
@@ -274,298 +424,395 @@ If there are any mis-matches between the values in `Y[;3]` and the types in `Y[;
...
 ```
 
-**Further Examples**
+This feature can be used to inject any raw text, although unless it is valid JSON it cannot then be re-imported.
 
-```apl
-      ⎕JSON ⊂ 3 ((↑[1]Items Price Qty)Fields)
-[{"Item":"Knife","Price":3,"Qty":23},{"Item":"Fork","Price":4,"Qty":45}]
+!!! Warning "Warning"
+    Out of convenience, it is common to initialise a list using a scalar rather than a one-element vector. Doing so for an eventual vector of character vectors will fail if the value is exported before a subsequent element is added:
```
FionaDyalog:
```suggestion
    It is common practice to initialise a list using a scalar rather than a one-element vector. Although this will succeed in generating a vector of character vectors, attempts to add a subsequent element will fail after the value is exported:
```

language-reference-guide/docs/system-functions/json.md
```diff
@@ -274,298 +424,395 @@ If there are any mis-matches between the values in `Y[;3]` and the types in `Y[;
...
+    ["foo"]
+    ```
 
-      ⎕JSON ⊂ 4 ((Items Price Qty)Fields)
-[{"Item":"Knife","Price":3,"Qty":23},{"Item":"Fork","Price":4,"Qty":45}]
-```
+<h4 class="example">Example</h4>
 
-Note that if you omit the wrapper, the operation generates a different result.
+The following illustrates how JavaScript objects can be exported; the object contains a JavaScript function which is specified by the contents of an enclosed character vector:
```
FionaDyalog:
```suggestion
This example illustrates how JavaScript objects can be exported; the object contains a JavaScript function that is specified by the contents of an enclosed character vector:
```

language-reference-guide/docs/system-functions/json.md
```diff
@@ -274,298 +424,395 @@ If there are any mis-matches between the values in `Y[;3]` and the types in `Y[;
...
-For codes 2, 3 and 4 the extension also provides the facility to optionally select elements of the dataset, so the array may contain 2, 3 or 4 items:
+Table: Wrapper codes { #wrapper-codes-table }
+
+| Wrapper code | Special array                                                                          | Advantage |
+|--------------|----------------------------------------------------------------------------------------|-----------|
+| `2`          | Single mixed-type matrix (first row is header vector)                                  | Preserves visual fidelity with a printed table |
+| `3`          | Two-element nested vector: value matrix and header vector                              | Allows indexing into the rows and columns of the data |
+| `4`          | Two-element nested vector: inverted table (vector of column vectors) and header vector | Less memory and faster lookups |
+
+For wrapper code `4`, if [**`HighRank`**](#variant-option-highrank) is `'Split'`, character columns can also be stored as character matrices rather than vectors of character vectors, providing even better performance, but note that `⎕JSON` will preserve trailing spaces.
```
FionaDyalog:
```suggestion
For wrapper code `4`, if [**`HighRank`**](#variant-option-highrank) is `'Split'`, character columns can also be stored as character matrices rather than vectors of character vectors, providing even better performance, but `⎕JSON` will preserve trailing spaces.
```

language-reference-guide/docs/system-functions/json.md
```diff
@@ -274,298 +424,395 @@ If there are any mis-matches between the values in `Y[;3]` and the types in `Y[;
...
 ```apl
-      1 ⎕JSON ⎕JSON'{"a": 1, "2a": 2}'
-{"a":1,"2a":2}
-
+      1 ⎕JSON⊂4(invertedTable header)(⊂⍬)(1 3)
+[{"item":"Knife","qty":23},{"item":"Fork","qty":45},{"item":"Spoon","qty":67}]
 ```
 
-Should you need to create and decode these names directly,`7162⌶` provides the same name mangling and un-mangling operations. See [JSON Translate Name](../primitive-operators/i-beam/json-translate-name.md).
+The following example selects the second record (Fork) and the first and third fields (`item` and `qty`):
```
FionaDyalog:
```suggestion
To select the second record (Fork) and the first and third fields (`item` and `qty`):
```

language-reference-guide/docs/system-functions/json.md
```diff
@@ -274,298 +424,395 @@ If there are any mis-matches between the values in `Y[;3]` and the types in `Y[;
...
+[{"item":"Fork","price":4,"qty":45}]
 ```
 
-When this JSON is imported as an APL namespace using `⎕JSON`, Dyalog converts the name *2a* to a valid APL name. The *name mangling* algorithm creates a name beginning with `⍙`.
-```apl
-      (⎕JSON'{"a": 1, "2a": 2}').⎕NL 2
-a  
-⍙2a
-```
+The following example selects the first and third fields (`item` and `qty`):
```
FionaDyalog:
```suggestion
To select the first and third fields (`item` and `qty`):
```

language-reference-guide/docs/system-functions/json.md
```diff
@@ -274,298 +424,395 @@ If there are any mis-matches between the values in `Y[;3]` and the types in `Y[;
...
+|--------------------|--------------------------------------------|
+| records            | `⊂code special recordIndices`              |
+| fields             | `⊂code special(⊂⍬)fieldIndices`            |
+| records and fields | `⊂code special recordIndices fieldIndices` |
 
-When Dyalog converts from JSON to APL data, and a member of a JSON object has a name which is not a valid APL name, it is renamed.
+<h4 class="example">Examples</h4>
 
-<h2 class="example">Example</h2>
+The following example selects the second record (Fork):
```
FionaDyalog:
```suggestion
To select the second record (Fork):
```

language-reference-guide/docs/system-functions/json.md
```diff
@@ -274,298 +424,395 @@ If there are any mis-matches between the values in `Y[;3]` and the types in `Y[;
...
 
-## JSON Name Mangling
+| Subset             | Wrapper form                               |
+|--------------------|--------------------------------------------|
+| records            | `⊂code special recordIndices`              |
+| fields             | `⊂code special(⊂⍬)fieldIndices`            |
+| records and fields | `⊂code special recordIndices fieldIndices` |
 
-When Dyalog converts from JSON to APL data, and a member of a JSON object has a name which is not a valid APL name, it is renamed.
+<h4 class="example">Examples</h4>
```
FionaDyalog, line 811: In these examples, is there a reason why "item" and "qty" are in APL font but "Fork" is shown in normal text font? If so, ignore this (but please explain it to me!)...if not, please make them all APL font.

FionaDyalog (changes requested):
Second tranche of feedback (now complete).

language-reference-guide/docs/system-functions/json.md
DyalogRichard: I don't like this proposed change. These are examples, not exercises. Therefore, the text should describe the effect of some change, not invite the reader to do it.

language-reference-guide/docs/system-functions/json.md
abrudz: Indeed.

PR #819, opened 2026-05-29
fix #818 table captions above

rikedyp:


abrudz (commented):
Can we get screen shots of the before and after?

abrudz (approved):

rikedyp:
**Before:**
image

**After**
image

PR #856, opened 2026-06-18
quadCSV added to Release Notes

FionaDyalog:
Reference #618


DyalogRichard (commented):
Two terminology issues, I think.

(1) There is no concept of differing variant options for monadic or dyadic uses. Variant creates a function with different default properties, which can then be used monadically or dyadically as usual. In this case it's just that ForceQuotes only has any effect when exporting.

(2) Data is not imported to the session; it is imported to the workspace.

DyalogRichard (approved):
LGTM

Issue #830, opened 2026-06-03
Rework the .NET Framework Interface Guide


FionaDyalog:
There are a few issues that ought to be resolved within the Dyalog for Microsoft Windows .NET Framework Interface Guide.

Specifically:

- [x] The compiler is referred to by a few different names (for example, the Dyalog .NET Compiler, Dyalog APLScript Compiler, the script compiler, the compiler, ...) The correct term should be applied throughout – The Dyalog .NET Compiler
- [ ] The examples should all be validated as they have not been checked for several releases. Ones that include mention of Dyalog versions earlier than v19.0 should be updated to refer to Dyalog v21.0 if possible.
- [ ] The application of code-style (APL vs nonAPL) should be checked and corrected where the incorrect one has been applied.
- [ ] Currently, examples with code blocks and line-by-line explanations are frequently broken up in to chunks of code followed by explanations of the lines within that chunk. Look at the instances of this, and determine whether it would be more user-friendly to have the entire code block followed by all the explanations
- [ ] Add cross-document links (for example, to functions, control structures, etc) throughout.
- [ ] Verify that all references to ".NET" are OK that way and should not be amended to ".NET Framework" – some incorrect ones might have crept in when sharing text with the .NET Interface Guide or just as shortcuts, and these should be updated.

FionaDyalog:
Note that this issue can't be worked on until #817 has been completed.

FionaDyalog:
Also see: #880

PR #857, opened 2026-06-18
"Compiler" references standardised

FionaDyalog:
Reference: #830


bear8642 (approved):
All looks good!

The commit (bc5664d5fd) does however include changes unrelated to the "Compiler" reference standardisation - these should be in separate commit(s).

Issue #849, opened 2026-06-16
The examples for Enclose should use boxed output


bear8642:
**Page on which the error exists**
[Enclose](https://docs.dyalog.com/21.0/language-reference-guide/primitive-functions/enclose/) 

**Version of the documentation in which you are reporting the error**
2026-06-12 main:4dcd72e6b9

**Describe the error**
The examples for enclose should use boxed output - perhaps via `]display` as [Enclose with Axes](https://docs.dyalog.com/21.0/language-reference-guide/primitive-functions/enclose/enclose-with-axes/) does?

**Additional information**
Both [Enclose with Axes](https://docs.dyalog.com/21.0/language-reference-guide/primitive-functions/enclose/enclose-with-axes/) and the [disambiguation page](https://docs.dyalog.com/21.0/language-reference-guide/symbols/left-shoe/) for `⊂` use boxed output in the examples

FionaDyalog:
Yes, that makes sense for enclose. The user command that we usually use for this is `]Boxing` (note capitalisation), so having a line at the start of the examples turning this on would be useful to explain the output..

bear8642:
See [Trains](https://docs.dyalog.com/20.0/programming-reference-guide/introduction/trains/#trains-of-trains) and [Behind](https://docs.dyalog.com/20.0/language-reference-guide/primitive-operators/behind/) for `]Boxing on` examples

PR #859, opened 2026-06-18
Use ]Boxing for enclose examples - v20.0

bear8642:
As #849 describes, enable ]Boxing on enclose examples for v20.0


FionaDyalog (approved):
LGTM

PR #862, opened 2026-06-22
Write .NET generics section in the .NET framework interface guide

pmikkelsen:


dotnet-framework-interface-guide/docs/dotnet-classes/advanced-techniques.md
```diff
@@ -218,3 +218,234 @@ For example, if <code class="language-nonAPL">MyClass</code> is a .NET class wit
 ```apl
       (⎕NEW ⍠ (⊂Double Int32)) MyClass (1 1)
 ```
+
+## Generics
+
+In .NET, a method, interface, or class can be _generic_, which means that it is a template or recipe for a _concrete_ method, interface, or class. What makes them generic is that they have a list of _type parameters_; the user must apply a matching number of _type arguments_ to create a concrete version. In the case of methods, it is not always necessary to apply type arguments, as the .NET interface can sometimes perform [type inference](#type-inference) to deduce the type arguments from the types of the method arguments.
+
+The syntax used to apply type arguments to methods, classes, and interfaces, is square brackets, for example `G[T]` where `G` is the generic entity, and `T` is a .NET type or a vector of .NET types. The types could be the result of applying types to a generic .NET class.
```
FionaDyalog:
```suggestion
The syntax used to apply type arguments to methods, classes, and interfaces is square brackets, for example, `G[T]`, where `G` is the generic entity and `T` is a .NET type or a vector of .NET types. The types could be the result of applying types to a generic .NET class.
```

dotnet-framework-interface-guide/docs/dotnet-classes/advanced-techniques.md
```diff
@@ -218,3 +218,234 @@ For example, if <code class="language-nonAPL">MyClass</code> is a .NET class wit
...
+LENGTH ERROR: No overload of the method expects the given number (10) of generic type arguments
+      ValueTuple.Create[10⍴Int32]
+                                ∧
+```
+
+#### Type Inference
+
+If the arguments to a generic method have a concrete .NET type, then their type information might be sufficient for the .NET bridge to unambiguously select a method overload and to automatically apply the needed type arguments. If there is any ambiguity about the type, such as when the arguments are regular APL arrays (for example the scalar `0`, which can be converted into a number of different .NET types), type inference will not take place. For example:
+
+```
```
FionaDyalog:
```suggestion
```apl
```

FionaDyalog (approved):
I have not checked this from a technical perspective. Approved on the assumption that suggested changes will be implemented.

Issue #784, opened 2026-05-07
00407 Completion of support for generics in .NET bridges phase I


FionaDyalog:
See project plan for details of changes needed.

Impacted documents:

- [x] Dyalog APL Language Reference Guide
- [x] .NET Interface Guide
- [x] .NET Framework Interface Guide

NOTE: The Dyalog v21.0 Release Notes will be updated for this under https://github.com/Dyalog/documentation/issues/618

pmikkelsen:
The issue was automatically closed when I merged the .NET interface guide changes, but there is still work to do, so I have reopened it.

PR #871, opened 2026-06-24
Changes for generics

FionaDyalog:
Reference: #618 (and #784 )

### DIFF
--- a/release-notes/docs/new-enhanced.md
+++ b/release-notes/docs/new-enhanced.md
@@ -22,7 +22,7 @@ The following system functions have been added:
 
 The following system functions have been enhanced:
 
-- [`⎕CSV`](https://docs.dyalog.com/21.0/language-reference-guide/system-functions/csv) – Comma Separated Values<br />A new variant option, **ForceQuotes**, has been added. This specifies when exported data has quotes around character/numeric fields.
+- [`⎕CSV`](https://docs.dyalog.com/21.0/language-reference-guide/system-functions/csv/) – Comma Separated Values<br />A new variant option, **ForceQuotes**, has been added. This specifies when exported data has quotes around character/numeric fields.
 
 ### I-beams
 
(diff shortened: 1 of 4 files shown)


release-notes/docs/new-enhanced.md
```diff
@@ -65,12 +70,14 @@ The following changes have been made to the TTY interface:
...
-
-### PCRE Library
+## Interfaces
 
-T.B.A.
-	
 ### .NET Interface
 
-T.B.A.
+Square brqcket (`[...]`) are now used to apply type arguments when instantiating generic methods, classes, and interfaces; this supersedes the I-beam that was used previously. For more information, see [Generics](https://docs.dyalog.com/21.0/net-interface-guide/dotnet-classes/advanced-techniques/#syntax).
```
pmikkelsen: brqcket -> bracket

release-notes/docs/new-enhanced.md
```diff
@@ -65,12 +70,14 @@ The following changes have been made to the TTY interface:
...
+
+### .NET Framework Interface
+
+In .NET, a method, interface, or class can be _generic_, which means that it is a template or recipe for a concrete method, interface, or class. What makes them generic is that they have a list of type parameters; the user must apply a matching number of type arguments to create a concrete version.
```
pmikkelsen: I think it would be simpler if we just said something like "The support for generics has been extended from only being supported in the .NET interface, to also be supported in the .NET framework interface", and then link to the generics section. Otherwise it looks a bit odd for someone reading the release notes from top to bottom, as they will first see a very short note about generics (new syntax, same functionality) for .NET interface, without any explanation about what it is. And then they will get a mini-introduction to what generics is in the .NET framework interface.

But if you think this is better, then I am OK with it.

dyavc (approved):

Issue #732, opened 2026-03-06
00425 .NET v10 support


FionaDyalog:
See project plan for details of changes needed.

Impacted documents:

- .NET Interface Guide

NOTE: The Dyalog v21.0 Release Notes will be updated for this under #618

PR #891, opened 2026-07-13
Version numbers and example upgrade process updated

FionaDyalog:
Reference: #732

### DIFF
--- a/dotnet-interface-guide/docs/installation.md
+++ b/dotnet-interface-guide/docs/installation.md
@@ -45,7 +45,7 @@ This is only an example of code that worked on a specific configuration in our t
 
 #### Upgrading .NET Support
 
-Dyalog v{{ version_majmin }} supports .NET v10.0 and later, but is configured to support .NET v10.0 by default. To support a later version of .NET, the following files need to be amended (this requires administrator rights) – in each case, the reference to "10.0" or "10.0.0" should be updated to the correct major version number:
+Dyalog v{{ version_majmin }} supports .NET v10.0 and later, but is configured to support .NET v10.0 by default. To support a later version of .NET, the following files need to be amended (this might require raised privileges) – in each case, the reference to "10.0" or "10.0.0" should be updated to the correct version number:
 
 - **[DYALOG]/Dyalog.Net.Bridge.deps.json**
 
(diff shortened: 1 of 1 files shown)


dotnet-interface-guide/docs/installation.md
```diff
@@ -33,45 +33,45 @@ On Raspberry Pi Bookworm, do not use the Microsoft-supplied <code class="languag
...
+sudo tar -zxvf /tmp/dotnet-runtime-10.0.0-linux-x64.tar.gz
 sudo  /usr/share/dotnet/dotnet /usr/bin/dotnet
 ```
 
 This is only an example of code that worked on a specific configuration in our tests; the latest instructions in Microsoft's .NET documentation should always be followed.
 
 #### Upgrading .NET Support
 
-Dyalog v{{ version_majmin }} supports .NET v8.0 and later, but is configured to support .NET v8.0 by default. To support a later version of .NET, the following files need to be amended (this requires administrator rights) – in each case, the reference to "8.0" or "8.0.0" should be updated to the correct major version number:
+Dyalog v{{ version_majmin }} supports .NET v10.0 and later, but is configured to support .NET v10.0 by default. To support a later version of .NET, the following files need to be amended (this requires administrator rights) – in each case, the reference to "10.0" or "10.0.0" should be updated to the correct major version number:
```
dyaandys: Remove "major"

dotnet-interface-guide/docs/installation.md
```diff
@@ -33,45 +33,45 @@ On Raspberry Pi Bookworm, do not use the Microsoft-supplied <code class="languag
...
-	    "tfm": "net8.0",  
+	    "tfm": "net10.0",  
 		"framework": {  
 		    "name": "Microsoft.NETCore.App",  
-			"version": "8.0.0"
+			"version": "10.0.0"
     ```
 
-The replacement version number can also be that of a .NET Release Candidate. For example, if you have downloaded .NET 10.0.0-rc.2, the version number in the aforementioned locations should be set to:
+The replacement version number can also be that of a .NET Release Candidate or Preview. For example, if you have downloaded .NET 11.0.0-rpreview.5, the version number in the aforementioned locations should be set to:
```
dyaandys: "if you have downloaded .NET 11.0.0-rpreview.5" - extraneous "r" before preview

dotnet-interface-guide/docs/installation.md
dyaandys: "the following files need to be amended (this requires administrator rights)" - I would change what's in parentheses to "(depending on the method used to install Dyalog, this may require raised privileges)"

dyaandys (changes requested):

dotnet-interface-guide/docs/installation.md
FionaDyalog: Going with "(this might require raised privileges)" unless you really want the previous bit...

dyaandys (approved):

PR #894, opened 2026-07-14
Fix #892 replacing 3+ line breaks with 2

### LINKED ISSUE #892: Remove extraneous line breaks

abrudz:
Also removes trailing blank lines


FionaDyalog (approved):
I have only verified a random selection, not the entire 1,278 files that have been changed. However, I am happy to approve this – if there are any oddities they can be caught and amended on an individual basis.

PR #906, opened 2026-07-21
Fix #750

### LINKED ISSUE #750: 00422 Allow ⎕UCS to accept/create UTF-8 in type 83 arrays
See project plan for details of changes needed.

Impacted documents:

- Dyalog APL Language Reference Guide

NOTE: The Dyalog v21.0 Release Notes will be updated for this under https://github.com/Dyalog/documentation/issues/618

abrudz:
Also lifts the bolded note-leader on UTF-16 to a proper heading to distinguish it from the new text about signed ints


language-reference-guide/docs/system-functions/ucs.md
FionaDyalog: X can also now be a 2-element nested array

FionaDyalog:
```suggestion
Monadic `⎕UCS` converts any character array to a numeric array of the same shape, or any numeric array to a character array of the same shape. When doing this, characters are converted to Unicode code points and Unicode code points are converted to characters.
```

FionaDyalog:
```suggestion
Dyadic `⎕UCS` translates between Unicode characters and one of three standard variable-length Unicode encoding schemes – UTF-8, UTF-16, or UTF-32. These represent a Unicode character string as a vector of 1-byte (UTF-8), 2-byte (UTF-16) and 4-byte (UTF-32) signed integer values respectively.
```

language-reference-guide/docs/system-functions/ucs.md
```diff
@@ -73,13 +73,29 @@ Dyadic `⎕UCS` is used to translate between Unicode characters and one of three
 947 949 953 945 32 963 959 965
 ```
 
-Because integers are *signed*, numbers greater than 127 will be represented as 2-byte integers (type 163), and are thus not suitable for writing directly to a native file. To write the above data to file, the easiest solution is to use  `⎕UCS` to convert the data to 1-byte characters and append this data to the file:
-```apl
+### UTF-8 Signed Integers
+
+Because integers are *signed*, numbers greater than 127 will be represented as 2-byte integers (type 163) which both double memory usage and are not suitable for writing directly to a native file without a costly conversion to 1-byte integers (type 83). In addition, reading such bytes from a native file can give negative numbers that are not suitable for direct consumption by `⎕UCS`. However, if `X` is `'UTF-8' 83`, `⎕UCS` will convert UTF-8 text by directly producing and consuming 1-byte integers (type 83) without passing through 2-byte integers. For example:
```
FionaDyalog: If useful to a user, explicitly include the detail that when a `[;2]` is `83` the values are emitted as type-83 bytes with values ¯128-127 (rather than 0-255).

FionaDyalog (changes requested):

language-reference-guide/docs/system-functions/ucs.md
abrudz: I don't think that'll help. If they don't know this already, it won't help here. The uninitiated user just needs to know that they need `83` to make things work.

language-reference-guide/docs/system-functions/ucs.md
```diff
@@ -11,21 +11,24 @@ search:
...
-- 'UTF-8'
-- 'UTF-16'
-- 'UTF-32'
+- `'UTF-8'`
+- `'UTF-8' 83`
+- `'UTF-16'`
+- `'UTF-32'`
 
-If not, a `DOMAIN ERROR` is issued.
+The character vector is the name of a variable-length Unicode encoding scheme and `83` specifies use of 1-byte integers (type 83) in result or right argument (whichever is numeric).
```
dyaandys: `83` specifies use of 1-byte signed integers (type 83)

Should we also say 
'0' specifies use of 1-byte unsigned integers (the default, can be omitted)

language-reference-guide/docs/system-functions/ucs.md
```diff
@@ -56,7 +59,7 @@ Unicode also contains the APL character set. For example:
 
 ## Dyadic `⎕UCS`
 
-Dyadic `⎕UCS` is used to translate between Unicode characters and one of three standard variable-length Unicode encoding schemes, UTF-8, UTF-16 and UTF-32. These represent a Unicode character string as a vector of 1-byte (UTF-8), 2-byte (UTF-16) and 4-byte (UTF-32) signed integer values respectively.
+Dyadic `⎕UCS` translates between Unicode characters and one of three standard variable-length Unicode encoding schemes – UTF-8, UTF-16, or UTF-32. These represent a Unicode character string as a vector of 1-byte (UTF-8), 2-byte (UTF-16) and 4-byte (UTF-32) signed integer values respectively.
```
dyaandys: unsigned integer values
```suggestion
Dyadic `⎕UCS` translates between Unicode characters and one of three standard variable-length Unicode encoding schemes – UTF-8, UTF-16, or UTF-32. These represent a Unicode character string as a vector of 1-byte (UTF-8), 2-byte (UTF-16) and 4-byte (UTF-32) integer values respectively. In the case of 1-byte integers, they can be specified to be signed or unsigned.
```

language-reference-guide/docs/system-functions/ucs.md
```diff
@@ -73,13 +76,29 @@ Dyadic `⎕UCS` is used to translate between Unicode characters and one of three
 947 949 953 945 32 963 959 965
 ```
 
-Because integers are *signed*, numbers greater than 127 will be represented as 2-byte integers (type 163), and are thus not suitable for writing directly to a native file. To write the above data to file, the easiest solution is to use  `⎕UCS` to convert the data to 1-byte characters and append this data to the file:
-```apl
+### UTF-8 Signed Integers
 
-      (⎕UCS 'UTF-8' ⎕UCS 'ABCÆØÅ') ⎕NAPPEND tn
+Because integers are *signed*, numbers greater than 127 will be represented as 2-byte integers (type 163) which both double memory usage and are not suitable for writing directly to a native file without a costly conversion to 1-byte integers (type 83). In addition, reading such bytes from a native file can give negative numbers that are not suitable for direct consumption by `⎕UCS`. However, if `X` is `'UTF-8' 83`, `⎕UCS` will convert UTF-8 text by directly producing and consuming 1-byte integers (type 83) without passing through 2-byte integers. For example:
```
dyaandys: how about something like

The short version:
⎕ucs consumes and returns unsigned integers.  For UTF-8 only, it is possible to pass and receive 1-byte signed integers by specifying the left argument to be ‘UTF-8’ 83.

The long version:
Internally Dyalog APL holds integer values as signed integers.  A 1-byte signed integer can contain values between ¯128 and 127, whereas a 1-byte unsigned integer can contain values between 0 and 255.
However, ⎕UCS was implemented such that numeric arguments and return values must be unsigned.  However, quite often numeric UTF-8 strings are brought into the interpreter or need to be passed out from the interpreter as signed integers (⎕nread and ⎕nappend are examples of system function which work with signed values).
Until version 21.0 it would have been necessary to convert such numeric strings from signed to unsigned and vice-versa.  This makes for more complex code, and the unsigned equivalent requires twice the space.    In 21.0, for 1-byte values only, it is possible, by extending the left argument to include 83 as the second element, so pass unsigned values to and from ⎕ucs.

language-reference-guide/docs/system-functions/ucs.md
```diff
@@ -11,21 +11,23 @@ search:
...
 
-- 'UTF-8'
-- 'UTF-16'
-- 'UTF-32'
+- `'UTF-8'`
+- `'UTF-16'`
+- `'UTF-32'`
 
-If not, a `DOMAIN ERROR` is issued.
+The second element is either `0` (the default) to consume and return byte values as positive integers or `83` to use 1-byte integers (type 83). `83` can currently only be used with `'UTF-8'`.
```
FionaDyalog:
```suggestion
The second element is either `0` (the default) to consume and return byte values as positive integers or `83` to use 1-byte integers (type 83). `83` can only be used with `'UTF-8'`.
```

language-reference-guide/docs/system-functions/ucs.md
```diff
@@ -73,13 +75,29 @@ Dyadic `⎕UCS` is used to translate between Unicode characters and one of three
 947 949 953 945 32 963 959 965
 ```
 
-Because integers are *signed*, numbers greater than 127 will be represented as 2-byte integers (type 163), and are thus not suitable for writing directly to a native file. To write the above data to file, the easiest solution is to use  `⎕UCS` to convert the data to 1-byte characters and append this data to the file:
-```apl
+### UTF-8 Signed Integers
 
-      (⎕UCS 'UTF-8' ⎕UCS 'ABCÆØÅ') ⎕NAPPEND tn
+By default, `⎕UCS` consumes and returns unsigned integers.  Numbers greater than 127 will be represented as 2-byte integers (type 163) which use twice as much memory and are not suitable for writing directly to a native file. In addition, reading such numbers from a native file will give negative numbers that are not suitable for direct consumption by `⎕UCS`. Conversion to and from 1-byte signed integers (type 83) is complicated and costly. For UTF-8 only, `⎕UCS` can directly produce and consume such integers with by using the left argument `'UTF-8' 83`. For example:
```
FionaDyalog:
```suggestion
By default, `⎕UCS` consumes and returns unsigned integers. Numbers greater than 127 will be represented as 2-byte integers (type 163); these use twice as much memory as 1-byte integers and are not suitable for writing directly to a native file. In addition, reading such numbers from a native file will give negative numbers that are not suitable for direct consumption by `⎕UCS`. Conversion to and from 1-byte signed integers (type 83) is complicated and costly. For UTF-8 only, `⎕UCS` can directly produce and consume such integers by using the left argument `'UTF-8' 83`. For example:
```

FionaDyalog: "reading such numbers from a native file will give negative numbers"

Not sure how to parse this...rewrite please. Do you mean "when converting from a native file, the numbers that are returned are negative" or something else?

language-reference-guide/docs/system-functions/ucs.md
```diff
@@ -73,13 +75,29 @@ Dyadic `⎕UCS` is used to translate between Unicode characters and one of three
...
+      tn←'letters.txt' ⎕NCREATE 0
+      ('UTF-8' 83 ⎕UCS 'ABCÆØÅ') ⎕NAPPEND tn
+      'UTF-8' 83 ⎕UCS ⎕NREAD tn 83 ¯1 0
+ABCÆØÅ
 ```
 
-**Note regarding UTF-16:** For most characters in the first plane of Unicode (0000-FFFF), UTF-16 and UCS-2 are identical. However, UTF-16 has the potential to encode all Unicode characters, by using more than 2 bytes for characters outside plane 1.
+### UTF-16 and UCS-2
+
+For most characters in the first plane of Unicode (0000-FFFF), UTF-16 and UCS-2 are identical. However, UTF-16 has the potential to encode all Unicode characters, by using more than 2 bytes for characters outside plane 1.
```
FionaDyalog:
```suggestion
For most characters in the [first plane of Unicode (0000-FFFF)](https://en.wikipedia.org/wiki/Plane_(Unicode)#Basic_Multilingual_Plane), UTF-16 and UCS-2 are identical. However, UTF-16 can potentially encode all Unicode characters by using more than 2 bytes for characters outside the first plane.
```

FionaDyalog (approved):
Approved on the assumptions that the feedback is incorporated.

language-reference-guide/docs/system-functions/ucs.md
```diff
@@ -11,21 +11,23 @@ search:
...
-- 'UTF-16'
-- 'UTF-32'
+- `'UTF-8'`
+- `'UTF-16'`
+- `'UTF-32'`
 
-If not, a `DOMAIN ERROR` is issued.
+The second element is either `0` (the default) to consume and return byte values as positive integers or `83` to use 1-byte integers (type 83). `83` can only be used with `'UTF-8'`.
+
+If `X` does not abide by the above restrictions, a `DOMAIN ERROR` is issued.
```
DyalogRichard, line 22: This description seems to preclude the previous format, which was that X is a simple character vector - to specify e.g. UTF-8, it says you now have to give the argument as <code>(,⊂'UTF-8')</code>. This is not the case - a simple character vector is still allowed.

DyalogRichard (changes requested):
The original text had some issues which made it at best unclear, at worst incorrect. In particular, it talked about UTF-8 being a vector of 1-byte characters, implying that you would get a 1-byte numeric vector when encoding text as UTF-8 - it did not explain why values in the range 0-255 actually occupy 2 bytes each in an APL array. 

This is not a fix for the old text, but the new addition is buillt upon it and does not really explain why encoding using signed values is _better_.  To do that, I think the underlying problem has to be fixed as well.

(There are related issues regarding encoding which do not have a similar knock-on problem - the text about UTF-16 implies there are some characters in the range U+0000 to U+FFFF which enode as more than two bytes ("For most characters in the first plane of Unicode (0000-FFFF) UTF-16 and UCS-2 are identical") and it states that UTF-32 is variable length, which it is not ("Three standard variable-length Unicode encoding schemes, – UTF-8, UTF-16 and, or UTF-32").

DyalogRichard (approved):

Issue #908, opened 2026-07-21
Unneccessary and ill-informed advice in the docs for ⎕UCS


DyalogRichard:
**Page on which the error exists**
https://docs.dyalog.com/20.0/language-reference-guide/system-functions/ucs/

**Version of the documentation in which you are reporting the error**
2026-05-06 v20.0:662a331439

**Describe the error**
The text "With a few exceptions, the first 256 Unicode code points correspond to the ANSI character set" is incorrect.

1. "ANSI" is ambiguous; this really means Windows-1252 (we have made this error elsewhere)
1. "Few" is poorly defined, but out of 256 codepoints, 32 are different - I would say that 1 in 8 is significantly more than "a few".

I don't think this sentence adds anything useful at all - just remove it.

PR #917, opened 2026-07-23
Unneccessary and ill-informed advice in the docs for ⎕UCS

bear8642:
As #908 describes, remove the bad sentence from `⎕UCS`


FionaDyalog (approved):
+1

Issue #854, opened 2026-06-17
Standardise examples to use ]Display or ]Boxing on


bear8642:
**Version of the documentation in which you are reporting the error**
2026-05-06 v20.0:662a331439

**Describe the error**
`]Boxing on` and `]Display` are both used throughout the docs.
We should standardise on using one or the other.

FionaDyalog:
In general, use `]Boxing`.

Some cases require `]Display` (for example, to decorate the border with axis/content information – use judgement).

Note the UpperCamelCase that is used for user commands (all capitals for group names) – see #469 

This impacts the following documents:

- [x] Dyalog APL Language Reference Guide
- [x] Dyalog Programming Reference Guide
- [x] Dyalog for Microsoft Windows Installation and Configuration Guide
- [x] Dyalog for Microsoft Windows UI Guide
- [x] Dyalog for Microsoft Windows Object Reference Guide
- [x] Dyalog for Microsoft Windows Interfaces Guide
- [x] Dyalog for Microsoft Windows .NET Framework Interface Guide
- [x] Dyalog for UNIX Installation and Configuration Guide
- [x] Dyalog for UNIX UI Guide

bear8642:
Also change `]Box` → `]Boxing` as comments on #867 identify

bear8642:
The use of `]Display` has been investigated, and left where it makes the most sense to occur.

PR #921, opened 2026-07-24
Standardise examples to use ]Display or ]Boxing ON

bear8642:
As #854 describes, make examples prefer to use ]Boxing instead of ]Display apart from when showing border information is useful.


FionaDyalog (approved):
+1

abrudz (approved):

Issue #754, opened 2026-04-10
00406.1 Improved thread scheduler - phase I


FionaDyalog:
See project plan for details of changes needed.

Impacted documents:

- Dyalog APL Language Reference Guide

NOTE: The Dyalog v21.0 Release Notes will be updated for this under https://github.com/Dyalog/documentation/issues/618

FionaDyalog:
Change of plan – this will now only be documented in the Dyalog v21.0 Release Notes.

PR #948, opened 2026-07-29
additions for thread scheduler changes

FionaDyalog:
Reference: #618 – main changes being done under #754


pmikkelsen (approved):
Looks good to me 👍

Issue #913, opened 2026-07-22
⎕ucs: rework information on signed vs unsigned integers


FionaDyalog:
The information for `⎕UCS` does not really cover everything that's needed with respect to signed vs unsigned integers, code points,, etc.

Proposal:

- A new section in the _Dyalog APL Programming Reference Guide_ covering signed vs unsigned, Unicode code points, etc
- Rework `⎕UCS` to remove extraneous information and instead point to the new section
- Update other system functions for which this is relevant (including cross-references to the new section). For example, `⎕NAPPEND` and `⎕NREAD`

dyaandys:
Richard and I will work on this together when he's back.

DyalogRichard:
Explaining why there are negative numbers here:

```apl
      83⎕DR'ABCü'
65 66 67 ¯4
```

but not here:

```apl
      ⎕UCS 'ABCü'
65 66 67 252
```

seems like a good starting point.

Other potential examples and counter-examples:

```apl
      1(219⌶)⍳10
 8 ¯55 1 0 0 0 0 10  ¯96 1 2 3 4 5 6 7 8 9 10 
```

```apl
      (4⍴256)⊤123456
0 1 226 64
```

dyaandys:
We'll stop at this point .. one day we might need to discuss signed/unsigned integers etc, but not now.

PR #972, opened 2026-08-04
913 rework text about bytes

dyaandys:
FWIW I'm not very keen on

If X does not abide by the above restrictions, a DOMAIN ERROR is issued.
at the top of ucs.md, but I'll defer to others

### DIFF
--- a/language-reference-guide/docs/system-functions/ucs.md
+++ b/language-reference-guide/docs/system-functions/ucs.md
@@ -53,7 +53,7 @@ Unicode also contains the APL character set. For example:
 
 ## Dyadic `⎕UCS`
 
-Dyadic `⎕UCS` translates between vectors of Unicode characters and one of three standard Unicode encoding schemes – UTF-8, UTF-16, or UTF-32. These represent a character vector as a vector of integers. In the case of UTF-8, the integers can be specified to be in the range `0` to `255` (when `X[2]` is `0`) or the range `¯128` to `+127`(when `X[2]` is `83`).
+Dyadic `⎕UCS` translates between vectors of Unicode characters and one of three standard Unicode encoding schemes – UTF-8, UTF-16, or UTF-32. These represent a character vector as a vector of integers. See the following section for details that are specific to UTF-8.
 ```apl
       'UTF-8' ⎕UCS 'ABC'
 65 66 67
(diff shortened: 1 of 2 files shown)


dyaandys:
Richard is happy with the text - he and I discussed it earlier this morning.

language-reference-guide/docs/system-functions/ucs.md
```diff
@@ -69,9 +75,9 @@ Dyadic `⎕UCS` translates between vectors of Unicode characters and one of thre
 947 949 953 945 32 963 959 965
 ```
 
-### UTF-8 Signed Integers
+### UTF-8 and Integer Ranges
 
-By default, `⎕UCS` uses unsigned integers. For UTF-8 only, if `X` is `'UTF-8' 83`, `⎕UCS` will instead use signed integers which are then represented as single bytes (type 83). For example:
+By default `⎕UCS` consumes and returns positive integers. In the case of `X` having the value `'UTF-8'` or `'UTF-8' 0`, `⎕UCS will consume and return integers in the range `0` to `255. In the case of `X` having the value `'UTF-8' 83`, `⎕UCS` will instead consume and return integers in the range `¯128` to `+127`. For example:
```
FionaDyalog, line 80:
```suggestion
By default, `⎕UCS` consumes and returns positive integers. When `X` has the value `'UTF-8'` or `'UTF-8' 0`, `⎕UCS` will consume and return integers in the range `0` to `255`. When `X` has the value `'UTF-8' 83`, `⎕UCS` will instead consume and return integers in the range `¯128` to `+127`. For example:
```

language-reference-guide/docs/system-functions/ucs.md
```diff
@@ -13,7 +13,7 @@ The optional left argument `X` is either a simple character vector or a one- or
 - `'UTF-16'`
 - `'UTF-32'`
 
-The second element (if present) is either `0` (the default) to consume and return byte values as positive integers or `83` to use 1-byte integers (type 83). `83` can only be used with `'UTF-8'`.
+The second element (if present) is either `0` (the default) which causes ⎕UCS to consume and return integers between 0 and 255 or `83` which causes ⎕UCS to consume and return integers between `¯128` and `+127`. `83` can only be used with `'UTF-8'`.  See [`⎕DR`](dr.md) for more information about type 83 values.
```
FionaDyalog, line 16:
```suggestion
The second element (if present) is either `0` (the default), which causes `⎕UCS` to consume and return integers between `0` and `255`, or `83`, which causes` ⎕UCS` to consume and return integers between `¯128` and `+127`. `83` can only be used with `'UTF-8'`.  See [`⎕DR`](dr.md) for more information about type 83 values.
```

language-reference-guide/docs/system-functions/ucs.md
```diff
@@ -81,7 +87,7 @@ By default, `⎕UCS` uses unsigned integers. For UTF-8 only, if `X` is `'UTF-8'
       'UTF-8' 83 ⎕UCS 'γεια σου'
 ¯50 ¯77 ¯50 ¯75 ¯50 ¯71 ¯50 ¯79 32 ¯49 ¯125 ¯50 ¯65 ¯49 ¯123
 ```
-This facilitates storing Unicode text in native files as UTF-8. For example:
+This facilitates storing Unicode text in native files as UTF-8 or being passed to or from `⎕NA` functions and in some cases will result in less workspace being needed to hold the integer vector argument or result. For example:
```
FionaDyalog, line 90:
```suggestion
This facilitates storing Unicode text in native files as UTF-8 or being passed to/from `⎕NA` functions. In some cases, this can result in a reduction in the amount of workspace needed to hold the integer vector argument or result. For example:
```

FionaDyalog (commented):
Approved once the changes have been implemented.

FionaDyalog:
Reference: #913

Issue #861, opened 2026-06-22
Minor issues in the Dyalog v21.0 documentation set


FionaDyalog:
THIS ISSUE IS A PLACEHOLDER FOR MINOR ISSUES SPOTTED/CORRECTED DURING THE DYALOG V21.0 DEVELOPMENT PHASE AND SHOULD REMAIN OPEN UNTIL DYALOG V21.0 GA – PLEASE DO NOT CLOSE IT WHEN THERE ARE NO OUTSTANDING ISSUES

- [x] /language-reference-guide/pcre-specifications/
Remove "now" from "The Dyalog interpreter now supports PCRE2"
- [x] /language-reference-guide/key-to-notation/
"function may refer to a primitive function, a system function, a defined (canonical, dfn or assigned) function or a derived (from an operator) function." -> "A _function_ can be a primitive function, a system function, a defined (canonical, dfn, or assigned) function or a derived (from an operator) function.
- [x] /language-reference-guide/primitive-operators/i-beam/monadic-operator-generator/
"Requires .NET or .NET framework" – framework -> Framework
- [x] /language-reference-guide/system-functions/json/#raw-text-wrapper
correct formatting error by inserting blank lines around text between examples
- [x] /language-reference-guide/system-functions/
Correct typos, remove duplicate rows, and amend incorrectly-categorised system functions
- [x] /unix-installation-and-configuration-guide/miscellaneous/
Page title should match what is in the ToC and the filename, currently do not match. Decide on what is most useful and apply in all three places.
- [x] /unix-installation-and-configuration-guide/starting-apl-from-scripts/
Page title should match what is in the ToC and the filename, currently do not match. Decide on what is most useful and apply in all three places.
- [x] /windows-ui-guide/browsing-net-classes/
ToC has ".Net" instead of ".NET" – needs to be corrected in the .yml file
- [x] /windows-installation-and-configuration-guide/runtime-applications-additional/
The last two headings on this page ("Either" and "Or") should not be headings! Change to be text (possibly with a slight wording tweak).
- [x] /language-reference-guide/system-functions/
Add `⎕SE` (name: Session Namespace, form: Reference)
- [x] /language-reference-guide/system-functions/system-functions-by-category/#built-in-objects-and-windows-gui
Add `⎕SE` (name: Session Namespace, form: Reference)
- [x] /windows-installation-and-configuration-guide/configuration-parameters/ride-init/
Italics have gone wrong and apparently missing text (probably markdown issue)
[...]

PR #980, opened 2026-08-06
Fix remaining minor documentation items

abrudz:
- dt: aim the datetime-code cross-reference at the Time numbers table id so it renders as a table reference instead of a blank
- ucs: reword the domain-error sentence to name the triggering case
- unix miscellaneous: title now matches the filename and contents list, with the session logfile text as a subsection
- unix starting APL from scripts: title, filename, and contents list now agree, and the redundant inner heading is gone

Part of #861.


FionaDyalog (approved):
LGTM

PR #960, opened 2026-07-30
Show the Documents folder name as it actually varies

abrudz:
The width is omitted for the 32-bit interpreter and the edition for the Classic edition, so the folder is Dyalog APL 19.0 Files for the 32-bit Classic edition, not Dyalog APL-32 19.0 Classic Files. Mark both parts optional instead of implying that each is always present, in the current guide and in the copy carried by the v19.0 release notes. While here, add the separator between the major and minor version numbers in the current guide, which the example already shows.

Refs Dyalog/documentation#959

PR #981, opened 2026-08-06
Correct the default log file path pattern

abrudz:
The Windows Documents folder is configurable, so the default log file path begins with <DocumentsDirectory> rather than a literal Users\<username>\Documents. The bit-width suffix appears only for the 64-bit interpreter, and only the Unicode edition carries the Unicode marker in the folder name.

Cf. #960


FionaDyalog (approved):
LGTM

PR #983, opened 2026-08-06
979 system documentation errors

### LINKED ISSUE #979: ⎕SYSTEM documentation errors
https://docs.dyalog.com/21.0/language-reference-guide/system-functions/system/

2026-08-04 main:bb74673d87

**Describe the error**
Review comments:

* R will remain serialisable using 1⎕JSON
  * This is syntactically invalid. Without a right argument it should be 1∘⎕JSON. In any case, do we want to make that promise? We don't know what we might want to add, and once we have ⎕APLAN, we might not consider it important to retain the ability to serialise using ⎕JSON.
* 64-bit operating systems can run 32-bit interpreters
  * This is not, in general, true. Most relevantly for us, 64-bit macOS cannot run 32-bit code (appears twice)
* Whether (1) or not (0) PCRE2 is in use is given by 10≤⎕SYSTEM.Features.PCRE.
  * This is not a useful hint - remove it. Every interpreter which supports ⎕SYSTEM contains PCRE2
* This can be converted to a character vector with ⎕UCS ⎕SYSTEM.OS.Newline.
  * No shit, Sherlock. An unneccessary admonition overload.
* A positive number means the local time zone is ahead of UTC. A negative number means behind UTC. For example, if ⎕OS.UTCOffset is 10, then when it is noon UTC (12:00), it is 10 PM local time (12+10, that is, 22:00).
  * Ditto. Whereas it might be worth saying something about DST (preferably not as an admonition). Also, change the description to "The number of hours by which the current local time zone is offset from UTC" (addition of "current").
* 1 0 1≡10 22000⍸⎕SYSTEM.OS.Version
  * This is wrong; 10.0.22000 is Windows 11 21H2. I believe that: 10.0.XXXXX indicates Windows 10 or Windows 11, and that XXXXX ≥ 22000 indicates Windows 11. But it's at Microsoft's whim and we shouldn't claim stuff we cannot be sure of.

abrudz:
Contentious points:

> * Whether (1) or not (0) PCRE2 is in use is given by 10≤⎕SYSTEM.Features.PCRE.
>   * This is not a useful hint - remove it. Every interpreter which supports ⎕SYSTEM contains PCRE2

There might be a PCRE3 in the future, and this teaches a pattern. Since we don't know what version number would indicate PCRE3, I've flipped the check to be for PCRE1.

> * This can be converted to a character vector with ⎕UCS ⎕SYSTEM.OS.Newline.
>   * No shit, Sherlock. An unneccessary admonition overload.

@FionaDyalog and @abrudz discussed it and settled on it being non-obvious to the uninitiated.

> * A positive number means the local time zone is ahead of UTC. A negative number means behind UTC. For example, if ⎕OS.UTCOffset is 10, then when it is noon UTC (12:00), it is 10 PM local time (12+10, that is, 22:00).
>   * Ditto. 

Ditto, but text simplified.

> * 1 0 1≡10 22000⍸⎕SYSTEM.OS.Version
>   * This is wrong; 10.0.22000 is Windows 11 21H2. I believe that: 10.0.XXXXX indicates Windows 10 or Windows 11, and that XXXXX ≥ 22000 indicates Windows 11. But it's at Microsoft's whim and we shouldn't claim stuff we cannot be sure of.

Well, nobody can ever use 10.0.22000, but changed to 21999. Also hardened to avoid false positives in case Microsoft restarts build numbering for a Windows 11+.

### DIFF
--- a/language-reference-guide/docs/system-functions/system.md
+++ b/language-reference-guide/docs/system-functions/system.md
@@ -29,13 +29,12 @@ Getting the currently running APL interpreter's version number:
 21
 ```
 
-The result `R` is a namespace in the root of the namespace where `⎕SYSTEM` was called (`#` or `⎕SE`) and contains only namespace members. Each sub-namespace contains only variable members.
+The result `R` is a namespace in the root (`#` or `⎕SE`) of the current namespace. `R` contains only namespace members, each of which only contains variable members.
 
-!!! Info "Information"
-    More members might be added in a future release of Dyalog, but `R` will remain serialisable using [`1∘⎕JSON`](json.md).
+More members might be added in a future release of Dyalog, but `R` will remain serialisable using [`1∘⎕JSON`](json.md).
 
 !!! Warning "Warning"
-    Do attempt to add or modify members as changes will not be persisted. Instead, clone the namespace using `⎕NS ⎕SYSTEM` and modify the result, but note that this locks down dynamic values like `⎕SYSTEM.Directories.Current` and `⎕SYSTEM.OS.UTCOffset`.
+    Do not attempt to add or modify members (changes will not be persisted); only modify clones of the namespace (`⎕NS ⎕SYSTEM`). Note that this locks down dynamic values like `⎕SYSTEM.Directories.Current` and `⎕SYSTEM.OS.UTCOffset`.
 
 ## Members of the Result
 
@@ -57,17 +56,15 @@ This namespace provides pertinent locations in the file system.
 #### Directories.Current
 The current working directory.
 
-!!! Hint "Hints and Recommendations"
-    This can be changed using the `]CD` user command, but doing so after [tying component files](ftie.md) or [native files](ntie.md) or [associating external functions](na.md) can lead to data loss.
-
 #### Directories.Initial
 The directory from which Dyalog was started.
 
 #### Directories.Temp
 The operating system's recommended location for temporary files.
 
-!!! Info "Information"
-    This location is for an individual user and can be cleaned up without warning, so use it only as a place to put files that will immediately be used and will not be needed later. It is good practice to [delete](ndelete.md) such files when no longer needed.
+This location is for an individual user and can be cleaned up without warning, so use it only as a place to put files that will immediately be used and will not be needed later.
+
+It is good practice to [delete](ndelete.md) such files when no longer needed.
 
 ### Executable
 This namespace provides information about the specific interpreter instance in which `⎕SYSTEM` was called.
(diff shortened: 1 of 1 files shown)


language-reference-guide/docs/system-functions/system.md
```diff
@@ -226,10 +226,10 @@ The character used to separate multiple filenames in a single character vector.
 The preferred file extension for [shared libraries](https://en.wikipedia.org/wiki/Shared_library) – one of `'.dll'` (Microsoft Windows), `'.dylib'` (macOS), or `'.so'` (Linux).
 
 #### OS.UTCOffset
-The number of hours by which the local time zone is offset from [UTC](https://en.wikipedia.org/wiki/Coordinated_Universal_Time).
+The number of hours by which the current local time zone (with [daylight saving time](https://en.wikipedia.org/wiki/Daylight_saving_time) taken into account) is offset from [UTC](https://en.wikipedia.org/wiki/Coordinated_Universal_Time).
 
 !!! Hint "Hints and Recommendations"
-    A positive number means the local time zone is ahead of UTC. A negative number means behind UTC. For example, if `⎕OS.UTCOffset` is `10`, then when it is noon UTC (12:00), it is 10 PM local time (`12+10`, that is, 22:00).
+    For example, if `⎕OS.UTCOffset` is `¯3`, then when it is 10:00 UTC, it is 07:00 local time (`10+¯3`).
```
FionaDyalog, line 232: I don't think this one needs to be a hint, just standard text.

language-reference-guide/docs/system-functions/system.md
```diff
@@ -239,7 +239,7 @@ A three-element integer vector:
 - AIX: version and release number, followed by a `0`
 
 !!! Warning "Warning"
-    Windows 11 identifies itself as "Windows 10". Whether (`1`) or not (`0`) Windows 10 is in use is given by `1 0 1≡10 22000⍸⎕SYSTEM.OS.Version`.
+    Windows 11 identifies itself as "Windows 10". Whether (`1`) or not (`0`) Windows 10 is in use is given by `1 0 2≡10 11 21999⍸⎕SYSTEM.OS.Version`.
```
FionaDyalog, line 242: I think this needs further clarification, as the first sentence implies that the second one could refer to Windows 10 or Windows 11.

FionaDyalog (commented):
Approved with proviso that the feedback is incorporated.

language-reference-guide/docs/system-functions/system.md
```diff
@@ -32,7 +32,7 @@ Getting the currently running APL interpreter's version number:
 The result `R` is a namespace in the root of the namespace where `⎕SYSTEM` was called (`#` or `⎕SE`) and contains only namespace members. Each sub-namespace contains only variable members.
 
 !!! Info "Information"
-    More members might be added in a future release of Dyalog, but `R` will remain serialisable using [`1⎕JSON`](json.md).
+    More members might be added in a future release of Dyalog, but `R` will remain serialisable using [`1∘⎕JSON`](json.md).
```
DyalogRichard, line 35: This still has the problem that it assumes it will remain serialisable, and there is every reason to suppose it might not.

language-reference-guide/docs/system-functions/system.md
```diff
@@ -152,7 +152,7 @@ Boolean indicating whether (`1`) or not (`0`) [Object Linking and Embedding](../
 Full version of the built-in [Perl Compatible Regular Expressions](../pcre-specifications) engine, for example `10 47`.
 
 !!! Hint "Hints and Recommendations"
-    Whether (`1`) or not (`0`) PCRE2 is in use is given by `10≤⎕SYSTEM.Features.PCRE`.
+    Whether (`1`) or not (`0`) PCRE1 is in use is given by `8≥⊃⎕SYSTEM.Features.PCRE`.
```
DyalogRichard, line 155: This can still never be true.

language-reference-guide/docs/system-functions/system.md
DyalogRichard: I don't think it is needed at all.

language-reference-guide/docs/system-functions/system.md
```diff
@@ -239,7 +239,7 @@ A three-element integer vector:
 - AIX: version and release number, followed by a `0`
 
 !!! Warning "Warning"
-    Windows 11 identifies itself as "Windows 10". Whether (`1`) or not (`0`) Windows 10 is in use is given by `1 0 1≡10 22000⍸⎕SYSTEM.OS.Version`.
+    Windows 11 identifies itself as "Windows 10". Whether (`1`) or not (`0`) Windows 10 is in use is given by `1 0 2≡10 12 21999⍸⎕SYSTEM.OS.Version`.
```
DyalogRichard, line 242: This is far too fragile, considering we don't know what Microsoft will do, nor any of the other OSes on which we run. Publishing this would be just out and out reckless.

DyalogRichard (dismissed):

DyalogRichard:
> @FionaDyalog and @abrudz discussed it and settled on it being non-obvious to the uninitiated.

I suggest that you use it in an example (thanks, Fiona, for the suggestion earlier). This is a Language Reference, not a User Guide. Hints and Recommendations (of which this is now full) do not belong in a LR.

language-reference-guide/docs/system-functions/system.md
abrudz: Rephrased.

language-reference-guide/docs/system-functions/system.md
abrudz: Removed.

language-reference-guide/docs/system-functions/system.md
abrudz: We changed some specs to ensure it would work, and we will make sure it is so going forward.

language-reference-guide/docs/system-functions/system.md
abrudz: This is [the official way to distinguish them](https://learn.microsoft.com/en-us/answers/questions/586619/windows-11-build-ver-is-still-10-0-22000-194).

Issue #581, opened 2025-10-22
Enhance the notes on primitive functions


FionaDyalog:
Re https://docs.dyalog.com/20.0/language-reference-guide/primitive-functions/notes/

This page needs to be reworked; check the validity of the information that's there and add additional sections such as:

- Conformability
- Fill Elements
- Axes
- Migration Level
- Implicit Arguments
- Singleton Extension

See the remaining Flare project for source information

PR #987, opened 2026-08-06
List what each system variable is an implicit argument of

### LINKED ISSUE #885: System functions: Add information to system functions that can be implicit arguments
Several system functions impact the behaviour of various primitive functions, primitive operators, and other functionality. There should be information on the pages relating to the following system functions about the things for which they are implicit arguments:

- [ ] ⎕CT 
- [ ] ⎕DCT
- [ ] ⎕DIV
- [ ] ⎕FR
- [ ] ⎕IO
- [ ] ⎕ML
- [ ] ⎕PP
- [ ] ⎕RL

This should be added in the following format on each of the impacted pages (ignore the horizontal lines, they are just there to identify the text in this issue!):

---------------------

⎕FOO is an [implicit argument](../primitive-functions/notes/#implicit-arguments) of:

- monadic functions: &lt;list with cross-references>
- dyadic functions: &lt;list with cross-references>
- operators: &lt;list with cross-references>
- system functions: &lt;list with cross-references>

---------------------

Note: Bullets from the above list should only be included if there is 1 or more things to go in the list. The lists for bullets 1 and 2 can be copied from [Table 2 in Implicit Arguments](https://docs.dyalog.com/21.0/language-reference-guide/primitive-functions/notes/#implicitargs). The lists for bullets 3 and 4 can be obtained from [Table 1 in Modifying Language Behaviour](https://docs.dyalog.com/20.0/language-reference-guide/system-functions/system-functions-by-category/#modifying-language-behaviour).

Ideally this information would be in a similar place on each page (near the top?), but there needs to be some flexibility here depending on the content of the page so I'm not going to dictate location beyond that.

abrudz:
Fixes #885.

Each of the ⎕CT, ⎕DCT, ⎕DIV, ⎕FR, ⎕IO, ⎕ML, ⎕PP, and ⎕RL pages now states, near the top, what it is an implicit argument of: a cross-referenced list of the monadic functions, dyadic functions, operators, system functions, and other syntax affected (each bullet present only when non-empty), sourced from the implicit-arguments tables in the notes on primitive functions and in the system functions by category.

Note: the ⎕ML entry uses the corrected dependence (monadic ∊ ↑ ⊃ ≡; dyadic ⊂; ⎕TC). The #implicitargs table in notes.md still lists monadic ⊂ ⊆ for ⎕ML and needs the same correction separately (#581).

### DIFF
--- a/language-reference-guide/docs/system-functions/ct.md
+++ b/language-reference-guide/docs/system-functions/ct.md
@@ -5,15 +5,22 @@ search:
 
 # <span>Comparison Tolerance</span> `⎕CT`
 
-The value of `⎕CT` determines the precision with which two numbers are judged to be equal.  Two numbers, `X` and `Y`, are judged to be equal if `(|X-Y)≤⎕CT×(|X)⌈|Y`where `≤` is applied without tolerance.
+The value of `⎕CT` determines the precision with which two numbers are judged to be equal.
+
+`⎕CT` is an [implicit argument](../primitive-functions/notes.md#implicit-arguments) of:
+
+- monadic functions: [`⌈`](../primitive-functions/ceiling.md), [`⌊`](../primitive-functions/floor.md), [`∪`](../primitive-functions/unique.md), [`≠`](../primitive-functions/unique-mask.md)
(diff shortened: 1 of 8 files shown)


FionaDyalog (commented):
Looks good. I have not checked the links.

Issue #988, opened 2026-08-10
00436 Substitution parameters in config settings


FionaDyalog:
Details of the changes needed are in the project plan.

Impacted documents:

- Dyalog for Unix Installation and Configuration Guide
- Dyalog for Microsoft Windows Installation and Configuration Guide
- Dyalog APL Language Reference Guide

The changes for the Dyalog v21.0 Release notes will be implemented under #618

PR #989, opened 2026-08-10
configuration setting information added

FionaDyalog:
Reference: #618 (also #988)

### DIFF
--- a/release-notes/docs/new-enhanced.md
+++ b/release-notes/docs/new-enhanced.md
@@ -2,8 +2,6 @@
 
 This page describes the changes and new features in Dyalog v21.0 compared with Dyalog v20.0.
 
-<p style="color:red;">This document is currently being developed and will not be finalised until nearer the release of Dyalog v21.0.</p>
-
 ## Language Changes
 
 ### System Functions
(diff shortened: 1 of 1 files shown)


release-notes/docs/new-enhanced.md
```diff
@@ -62,6 +62,13 @@ This specifies whether support for external variables is enabled.
 - [`LAYOUT_FILE`](https://docs.dyalog.com/21.0/windows-installation-and-configuration-guide/docs/configuration-parameters/layout-file/) (Microsoft Windows only)  
 This specifies the path (absolute or relative to the working directory) and name of the Session layout file.
 
+### Configuration Settings
+
+The following changes have been made to configuration settings:
+
+- Special, pre-defined, substitutions can now be made within any configuration settings; previously this only worked within configuration files. It remains unsupported beyond configuration settings, for example, `)LOAD [DYALOG]/myws.dws` is not valid. There is a small chance that existing configuration settings could now see unintended substitution; to ensure that this does not happen, when including literal square brackets in a string, the `[` should be prefixed with a backslash, that is `/[` (the backslash is removed from the subsequent value).
```
DyalogRichard: Looks good, except I think "pre-defined" is a bit misleading. Any configuration setting can be substituted into another. "pre-defined" may inadvertantly imply (a) there is some specific set of things that can be substituted (in the case of special names beginning with = this is true, but that's a distinct section below); (b) there is no ordering requirement: config setting A can have B substituted into it even if B is defined later on.

DyalogRichard (changes requested):

PR #1002, opened 2026-08-11
Fix #915

### LINKED ISSUE #915: Show an example of when ⎕VGET 1 (or ¯1) might be useful
Re: https://docs.dyalog.com/20.0/language-reference-guide/system-functions/vget/

2026-05-06 v20.0:662a331439

I (Andy) wondered why ⎕vget 1 was valid.  Adam gave the example of
<pre>
     ∇ foo
[1]   l1:
[2]   l2:
[3]    ⎕VGET ¯1
     ∇
      foo
  l1  1   l2  2
</pre>

which may not be that useful, but what might be more useful is
<pre>
     ∇ foo
[1]   l1:
[2]   l2:
[3]    ∘
     ∇
      foo
SYNTAX ERROR
foo[3] ∘
       ∧
      ⎕vget ¯1
  l1  1   l2  2
</pre>
That is, you can check what labels are in the current function when the function is suspended (or you trace into the function and call ⎕vget ¯1

abrudz:


FionaDyalog (approved):
The example is not the same as was initially suggested, but as long as it illustrates the same point then LGTM.

Issue #850, opened 2026-06-17
Enhance the interlinking/cross-referencing across the entire documentation set


FionaDyalog:
There are multiple places in the documentation that refer to features that are documented elsewhere, either within the same document or within other documents (for example, references to primitive functions). Adding cross-references to such features would enhance the usefulness of our documentation.

As a rule of thumb, the first time a feature is mentioned on a page it would be good to add a cross-reference (subsequent mentions on the same page do not need this) unless it is within a block of code – common sense applies! 

This impacts the following documents:

- [ ] Dyalog APL Language Reference Guide
- [ ] Dyalog Programming Reference Guide
- [ ] Dyalog for Microsoft Windows Installation and Configuration Guide
- [ ] Dyalog for Microsoft Windows UI Guide
- [ ] Dyalog for Microsoft Windows Object Reference Guide
- [ ] Dyalog for Microsoft Windows Interfaces Guide
- [ ] Dyalog for Microsoft Windows .NET Framework Interface Guide
- [ ] Dyalog for UNIX Installation and Configuration Guide
- [ ] Dyalog for UNIX UI Guide
- [ ] Shared Code Files User Guide
- [ ] Compiler User Guide
- [ ] .NET Interface Guide

Any new documents brought across from the previous project should also be checked/amended as appropriate.

PR #1006, opened 2026-08-11
Show a ⎕USING example with a user-written C# class (#802)

### LINKED ISSUE #802: Document ⎕USING example with external C# class module
**Page on which the enhancement should be made**
.NET Interface Guide - Section 3

**Describe the enhancement you would like**
The .NET interface documentation seems to only show how to work with the supplied System .NET modules as examples. It would be helpful to show an example of ⎕USING with an external C# module (either from NuGET or a very basic C# class module built by the user). This was something I spent a bit trying to figure out that I think would be beneficial to showcase.

**Additional information**
I have a very basic module working to demonstrate this. If time permits, I can share it for reference?

abrudz:
Add a Using a User-Written Class section to Using .NET Classes in both the .NET Interface Guide and the .NET Framework Interface Guide: a small C# class compiled to an assembly, then loaded through ⎕USING by its .NET namespace and file path and called. Example contributed by the reporter.

Fixes #802

### DIFF
--- a/dotnet-framework-interface-guide/docs/dotnet-classes/using-dotnet-classes.md
+++ b/dotnet-framework-interface-guide/docs/dotnet-classes/using-dotnet-classes.md
@@ -42,6 +42,29 @@ This assumes that `⎕USING` has been defined correctly – there must be a refe
 
 The rest of this page describes the mechanism by which Dyalog associates the class name with a class in a .NET namespace.
 
+## Using a User-Written Class
+
+The classes used above are supplied with .NET, but [`⎕USING`](../../../language-reference-guide/system-functions/using/) can also load your own compiled .NET assemblies. For example, suppose the following C# class is compiled into an assembly file named `testlib.dll`:
+```C#
+namespace Foobar
+{
+    public static class Adder
+    {
+        public static int AddInts(int x, int y)
+        {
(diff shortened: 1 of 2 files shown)


johndaintree (approved):

dotnet-framework-interface-guide/docs/dotnet-classes/using-dotnet-classes.md
```diff
@@ -42,6 +42,29 @@ This assumes that `⎕USING` has been defined correctly – there must be a refe
 
 The rest of this page describes the mechanism by which Dyalog associates the class name with a class in a .NET namespace.
 
+## Using a User-written Class
+
+The classes used in the previous examples are supplied with .NET, but [`⎕USING`](../../../language-reference-guide/system-functions/using/) can also load your own compiled .NET assemblies. For example, suppose the following C# class is compiled into an assembly file named **testlib.dll**:
```
FionaDyalog: Remove link to ⎕USING – it is not the first mention on the page. When the documentation set is enhanced with additional interlinking (see #850) a link should be added to an earlier reference on this page.

FionaDyalog (approved):
Approved with one amendment

Issue #887, opened 2026-07-09
System functions: Standardise references to variants


FionaDyalog:
Several system functions allow variant options, but they are not presented consistently.

Variant options should be documented as APL code (currently they are mostly in bold body font).

The following standard for structure should be followed:

- Variant options should be introduced with a heading "Variant Options".
  - This heading should be followed by summary information of what the variant options are, possibly as a table if there are multiple variant options. Default values and the principal variant option should be mentioned.
- Beneath the main headings, each variant option should have a title "Variant Option: Foo".
  - This heading should be followed by a description of what that specific variant option does, including examples.

For an example of this structure, see /language-reference-guide/system-functions/json/#variant-options (Dyalog v21.0 onwards).

Note that this structure is the ideal standard; as ever, common sense should apply, and usefulness to the end reader must take priority.

Variant options can currently be set for the following system functions – these need to be updated in accordance with the above standard:

- [ ] `⎕CSV`
- [ ] `⎕DT`
- [ ] `⎕ED`
- [ ] `⎕FCHK`
- [ ] `⎕FCOPY`
- [ ] `⎕FCREATE`
- [ ] `⎕⎕FIX`
- [ ] `⎕FSTIE`
- [ ] `⎕FTIE`
- [ ] `⎕JSON`
- [ ] `⎕MKDIR`
- [ ] `⎕NCOPY`
- [ ] `⎕NCREATE`
- [ ] `⎕NDELETE`
- [ ] `⎕NEXISTS`
- [ ] `⎕NINFO`
- [ ] `⎕NMOVE`
- [ ] `⎕NPUT`
- [ ] `⎕NS`
- [ ] `⎕SHELL`
- [ ] `⎕VSET`
- [ ] `⎕XML`

The [Variant](https://docs.dyalog.com/20.0/language-reference-guide/primitive-operators/variant/) page should not list the system functions that it can apply to – although this might be nice to have, it has been deemed an unnecessary synchronisation risk. That decision can always be revisited once this work has been completed.

PR #1009, opened 2026-08-11
Clarify when ⎕ED returns (#805)

### LINKED ISSUE #805: ⎕ED returns immediately if called from session
https://docs.dyalog.com/20.0/language-reference-guide/system-functions/ed/

2026-05-06 v20.0:662a331439

**Describe the error**
Documentation for ⎕ED incorrectly states that it returns a result only when the user closes all windows associated with the edit call; this is true only if called from a defined function or operator: if called from the session, the result is returned immediately

**Additional information**
[From Mantis](https://mantis.dyalog.com/view.php?id=8413)

abrudz:
⎕ED returns a result only when all windows are closed when it is called from a defined function or operator. Called from the Session it opens the Edit windows and returns a null result immediately.

The two ⎕ED-behaviour paragraphs also carry the may-to-can/might wording from the variant-options PR (#887), so the shared region is identical on both branches and the two PRs do not conflict.

Fixes #805

### DIFF
--- a/language-reference-guide/docs/system-functions/ed.md
+++ b/language-reference-guide/docs/system-functions/ed.md
@@ -24,11 +24,11 @@ If `Y` names an existing object, the type specification for that name in `X` is 
 If `X` is `⋄`, `Y` must be undefined or an array.
 The Editor opens in array-notation mode; the resulting array can be of any type or structure.
 
-If `⎕ED` is called from the Session, it opens Edit windows for the object(s) named in `Y` and returns a null result.  The cursor is positioned in the first of the Edit windows opened by `⎕ED`, but may be moved to the Session or to any other window which is currently open.  The effect is almost identical to using `)ED`.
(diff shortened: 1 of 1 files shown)


language-reference-guide/docs/system-functions/ed.md
```diff
@@ -24,11 +24,11 @@ If `Y` names an existing object, the type specification for that name in `X` is
...
+If `⎕ED` is called from a defined function or operator, its behaviour is different. On asynchronous terminals, the Edit windows are automatically displayed in "full-screen" mode (ZOOMED). In all implementations, the user is restricted to those windows named in `Y`. The user cannot skip to the Session even though the Session might be visible.
 
-`⎕ED` terminates and returns a result ONLY when the user explicitly closes all the windows for the named objects. In this case the result contains the names of any objects which have been newly (re)fixed in the workspace as a result of the `⎕ED`, and has the same structure as `Y`.
+In this case, `⎕ED` terminates and returns a result only when the user explicitly closes all the windows for the named objects. The result contains the names of any objects that have been newly (re)fixed in the workspace as a result of the `⎕ED`, and has the same structure as `Y`.
```
FionaDyalog: Add this to the end of the previous paragraph rather than having it in a new paragraph to emphasise the continuation; don't then need "in this case".

FionaDyalog (changes requested):

FionaDyalog (approved):

PR #1010, opened 2026-08-11
758 step updown

### LINKED ISSUE #758: Step property text does not describe what it does for UpDown/Spinner/ProgressBar/TrackBar objects
https://docs.dyalog.com/20.0/object-reference/properties/step/

2026-04-07 main:8267b09753

**Describe the error**
content does not describe what happens for an UpDown or a Spinner object

**Additional information**
reported by a customer

abrudz:
Fixes #758


FionaDyalog (approved):
LGTM

Approved on the condition that this change has also been technically reviewed by someone appropriate.

johndaintree (approved):
That looks fine and doesn't need changing.  My only comment (FYI only I suppose) is for the UpDown and Spinner objects, we get a notification from Windows (as per https://learn.microsoft.com/en-us/windows/win32/api/commctrl/ns-commctrl-nmupdown) This includes a "signed integer value that represents the proposed change in the up-down control's position". We have observed that this is always +/- 1, which we multiply by STEP to get the adjustment. In theory there may be a UI that causes values other than +/- 1, but I've never seen it.

PR #1016, opened 2026-08-11
Fold the Editor and Tracer sub-pages into single pages (#108)

### LINKED ISSUE #108: Dyalog for Microsoft Windows UI Guide: Heading hierarchy not optimal
The sections on the Editor and the Tracer have multiple sub-sections. Currently some of these are seperate pages and some are sections on the main page. The division seems to be a tad arbitrary. I suggest making all the separate pages into sections on the main page so that the left-hand ToC has the structure:
...
Object Properties Dialog Box
The Editor
The Tracer
The Threads Tool
...
(Note I have tweaked some of these slightly for accuracy/consistency; if this complicates things through links then I can raise that as a separate issue).

abrudz:
Make Find and Replace Dialogs, Editing Scripts and Text Files, and Source as Typed sections of The Editor, and Inline Tracing a section of The Tracer, so the navigation reads Object Properties Dialog Box, The Editor, The Tracer, The Threads Tool. Repoint the affected inbound links to the new section anchors and rename the two nav entries for consistency.

Fixes #108


windows-ui-guide/mkdocs.yml
```diff
@@ -81,14 +81,9 @@ nav:
...
-  - Editor: editor.md
-  - Find and Replace Dialogs: find-and-replace-dialogs.md
-  - Editing Scripts and Text Files: editing-scripts-and-text-files.md
-  - Source as Typed: source-as-typed.md
-  - The Tracer: 
-      - tracer/index.md
-      - Inline Tracing: tracer/inline-tracing.md
+  - Object Properties Dialog Box: object-properties-dialog.md
+  - The Editor: editor.md
+  - The Tracer: tracer/index.md
```
FionaDyalog: As this is just a single file now, does not need to be "nested"...change from tracer/index.md -> tracer.md

FionaDyalog (changes requested):

FionaDyalog (approved):
I believe the only change since my previous review is the flattening of the structure, in which case this is approved.

